PR #47, opened 2015-11-11
Media Library plugin for CKEditor

phenaproxima:
This PR provides the _very_ early alpha of a simple Media Library plugin for CKEditor.

The media library is opened in a modal dialog box by a button in the CKE toolbar, and rendered client-side using Backbone. The items in the media library are fetched over AJAX using a view (which in turn causes the lightning_media module to rely on the core REST and Serialization modules), and when clicked they are directly embedded into the CKEditor document using Entity Embed. It just works.

https://github.com/drupal-media/entity_embed/pull/189 is related to this here PR. I don't think it's an outright dependency, but this work resulted in that patch.

Right now, the media library button will need to be added manually to a CKEditor configuration (and entity embedding will need to be enabled for the text format). Later we can maybe add that as a default to (certain? all?) text formats included with Lightning.


modules/lightning_features/lightning_media/lightning_media.info.yml
```diff
@@ -1,10 +1,9 @@
 name: Lightning Media
 type: module
-description: 'Provides install and configuration for great media authoring experiences.'
+description: Slick media handling for Lightning. So cool you'll make the A/C jealous.
```
balsama, line 3: ++

modules/lightning_features/lightning_media/lightning_media.info.yml
```diff
@@ -1,10 +1,9 @@
...
 core: 8.x
-package: lightning
+package: Lightning
 dependencies:
-  - embed
   - entity_embed
-  - entity_browser
-  - file_browser
+  - rest
+  - serialization
```
balsama, line 17: Need a dependency on media_entity here I think. Can you add that?

modules/lightning_features/lightning_media/lightning_media.info.yml
```diff
@@ -1,10 +1,10 @@
...
-package: lightning
+package: Lightning
 dependencies:
-  - embed
   - entity_embed
-  - entity_browser
-  - file_browser
+  - media_entity
+  - rest
+  - serialization
```
balsama, line 17: As discussed, I think all of the media entity plugins need to be requirements here. Please add:
- media_pinkeye
- media_entity_slideshow
- media_entity_instagram
- media_entity_twitter
- media_entity_image

modules/lightning_features/lightning_media/src/Plugin/EmbedType/MediaLibrary.php
```diff
@@ -0,0 +1,72 @@
...
+use Drupal\embed\EmbedType\EmbedTypeBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * @EmbedType(
+ *   id = "media_library",
+ *   label = @Translation("Media Library")
+ * )
+ */
+class MediaLibrary extends EmbedTypeBase implements ContainerFactoryPluginInterface {
```
davereid, line 22: I'm curious why this need to define a new embed type when you're basically embedding entities (which should be supported by Entity Embed, and adding a button to embed media_entity entities)?

phenaproxima, line 22: I think this is just future-proofing on my part so that we have extra flexibility if we need it. But yes, at the moment we're not doing anything dramatically different from what Entity Embed already does.

PR #50, opened 2015-11-25
Media Library widget for CKEditor

phenaproxima:
The finished (more or less) media library widget for CKEditor! It just effing works, right out of the box.


modules/lightning_features/lightning_media/config/install/core.entity_view_display.media.image.embedded.yml
```diff
@@ -0,0 +1,32 @@
+uuid: 52579bba-569e-497d-9400-b3fe7c6a752c
```
balsama, line 1: We don't want the uuid here, do we? It does make me think that we might be able to leverage UUIDs in our config files in whatever our long term plan is for updates though. hmmm.

modules/lightning_features/lightning_media/sass/library.scss
```diff
@@ -0,0 +1,91 @@
...
+    .ui-tabs-nav {
+      padding: 0;
+    }
+
+    .ui-tabs-panel {
+      @include clearfix();
+    }
+  }
+
+  #library {
```
balsama, line 19: Can we target something other than an ID here?

modules/lightning_features/lightning_media/sass/library.scss
```diff
@@ -0,0 +1,91 @@
...
+
+        img {
+          max-width: 100%;
+          height: auto;
+        }
+      }
+    }
+  }
+
+  #upload footer {
```
balsama, line 87: Another ID. Wondering if we can target something less specific.

PR #56, opened 2015-12-01
Sane default configuration for lightning_media

phenaproxima:
This PR adds a few very, very basic Behat tests to Lighting Media, and a whole lot of sane default configuration. That includes two new view modes for media entities -- embedded and media_library. The embedded view mode is used when embedding a media entity in CKEditor; media_library is not used at the moment, but the intention is that it will eventually be the view mode used in the media library dialog box exposed by CKEditor. (This depends on the ability to embed a rendered entity as a field in a view, which does not appear to be something that core gives us. I know it was possible in D7 -- possibly with a contrib module -- but I haven't yet found a solution for D8.)

The tests will NOT pass on Travis due to build issues -- namely, our make files are not feature branch-aware. I can explain how to run them manually if you want to test before merging.


phenaproxima:
Oh, and remember to squash the commits when merging :)

PR #78, opened 2016-01-05
Moving custom step definitions to lightning subcontext.

grasmash:


grasmash:
This change allows Lightning Behat tests to be run outside of the lightning context. I.E., via a different FeatureContext.php at a higher level in the project.

grasmash:
Don't merge... this isn't working, but it's not causing tests to fail?

grasmash:
Good, failing as expected now :)

grasmash:
@phenaproxima As with #79 and #80, this appears to be failing due to an upstream problem with the Behat test for the media feature.

phenaproxima:
The tests are passing in the main branch as of my last commit, so it seems to me that this is probably failing due to the changes made in behat.yml...?

grasmash:
@balsama So did the tests for this end up passing?

balsama:
Yeah, I didn't have any trouble locally and travis seems happy.

I've seen intermittent problems with one of the media tests on travis. Restarting the build from travis's UI has always resolved it and I haven't dug any deeper. Maybe you guys were seeing that?

PR #79, opened 2016-01-05
Adding more caching to Travis builds to decrease build time.

grasmash:


grasmash:
This failure seems unrelated to changes, re-run?

grasmash:
Added a commit to force a rerun.

grasmash:
@phenaproxima Yeah, this is getting the same unrelated Behat failure for the media feature that #80 is getting.

balsama:
Thanks @grasmash! Made us realize that we're not using those ssh hosts anymore too, but good to know that there is a plugin for that. Thanks!

Issue #94, opened 2016-01-22
Define dependencies in composer.json


grasmash:
We're currently using composer to manage dependencies as per the model set forth in github.com/drupal-composer/drupal-project.

However, using this does not allow Lightning to pull in its dependencies. 

Here's a headstart:
````
{
    "name": "acquia/lightning",
    "description": "The intent of Lightning is to provide a lightweight framework for building working solutions in Drupal.",
    "type": "drupal-profile",
    "repositories": [
        {
            "type": "composer",
            "url": "https://packagist.drupal-composer.org"
        }
    ],
    "require": {
        "composer/installers": "^1.0.20",
        "cweagans/composer-patches": "~1.0",
        "drupal/embed": "8.0-1.0#26265b5437bb0f1dcb916a6a1169d53382a23646",
        "drupal/entity_embed": "8.0-1.0#9760c0a790bccb30d7e15815fdfa6d1794453d3a",
        "drupal/media_entity": "8.0-1.0#f8decb81965037b05c51ce8c16eae8c37256e7a0",
        "drupal/media_entity_embeddable_video": "8.0-1.0#f7e51c2cccd875e1fd09d018c11fd50c93a71507",
        "drupal/media_entity_instagram": "8.0-1.0#75e41cbae35f260fa0e3e76f90562e26b4dc162d",
        "drupal/media_entity_twitter": "8.0-1.0#179e19e159f7df2172316030fc7c110b6f0d3efd",
        "drupal/media_entity_image": "8.0-1.0#f17384aeba3c8c00bbc57bfa8fa68529992873ce",
        "drupal/ctools": "8.0-3.0#e27c2663027a4f90dffb39f7fd6ed8a92e5d61d4",
        "drupal/layout_plugin": "8.0-1.0#993c829961b5b304a0094d7cd33f014edccf859e",
        "drupal/panels": "8.0-3.0#a458d139e2e9e35c37048cb30dc682087593bd4d",
        "drupal/page_manager": "8.0-1.0#3f15c87fec006e78954bb3e984cb80afeea4f5b2",
        "drupal/acquia_connector": "8.0-1.0#302fc469eb860892fc90403b5d758eb50a715f77",
        "drupal/metatag": "1.0-beta3",
        "drupal/core": "8.*"
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "extra": {
        "patches": {
            "drupal/ctools": {
[...]

PR #95, opened 2016-01-25
Adding composer and bower dependency definitions.

grasmash:


grasmash:
This is a work in progress. Still have not confirmed working functionality. Likely need to do some work with autoloader.

grasmash:
So, there are some interesting consequences of using Bower for front end dependency management:
- Build process is more lightweight. For dependencies, we typically pull in only the "dist" files rather than the entire library.
- The library paths change. We can keep things in /libraries, but the subdirectories change depending on the dependency's "main" value in its bower.json.

composer.json
```diff
@@ -0,0 +1,38 @@
...
+    "drupal/page_manager": "8.1.x-dev#3f15c87fec006e78954bb3e984cb80afeea4f5b2",
+    "drupal/acquia_connector": "8.1.x-dev#302fc469eb860892fc90403b5d758eb50a715f77",
+    "drupal/metatag": "8.1.0-beta3",
+    "drupal/core": "8.*"
+  },
+  "minimum-stability": "dev",
+  "prefer-stable": true,
+  "extra": {
+    "patches": {
+      "drupal/ctools": {
```
danepowell, line 33: doesn't media_entity_twitter also have a patch? https://github.com/acquia/lightning/blob/8.x-1.x/drupal-org.make#L33

balsama:
What do you think about this @phenaproxima ? Could we update our prepmake script to automatically generate the composer file too?

grasmash:
@balsama I've created a workaround for this issue on Bolt, so it's not as urgent. I've linked to a number of solutions in progress in #94, perhaps we should see if those move forward?

PR #96, opened 2016-01-27
Added bash scripts to build a default local install.

damontgomery:
I'm not sure if this fits how the project is set up, but this PR is intended to address the following concern.

When downloading the repo, setting up a site is not as simple as running `drush make`.

The travis file contains more advanced steps for the development build. Then there is also the normal Drupal set-up that you have to do for a local site.

For this reason, I've created two scripts,
`build.sh` to do what Travis does.  
`build-local-settings.sh` to do all the normal Drupal stuff.

I think there is value in `build.sh` because of the issue stated above. I think there may be value in the `build-local-settings.sh` because Lightning may be the entry point for a lot of people and knowing to create / copy those files is not the most straightforward thing in the world. This script is easy to read, commented, and will get someone set up in minutes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /tests/bin/
 vendor/
 node_modules/
+/docroot
--- a/README.md
+++ b/README.md
@@ -91,3 +91,50 @@ of site builds will use.
     # From /profiles/lightning/modules/lightning_features/lightning_media/tests/js;
     npm install && npm test
 
+## Quick install
+
+Run `./build.sh` to use drush make to build the distribution and copy the most up to date versions of the lightning modules into the site.
+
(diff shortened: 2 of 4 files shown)


balsama:
Thanks for this. One concern:

> When downloading the repo, setting up a site is not as simple as running drush make.

But it is! All that extra stuff is a workaround to get travis to test dev code. We don't necessarily want users building from dev code. Drush make will build the same thing as the latest tagged release on D.O, which is what we want people to be using.

I like the idea of a quick-start script, but I don't want to encourage people to be build from dev code.

damontgomery:
OK, makes sense. The local build file is not specific to lightning, so probably makes sense to close this.

PR #103, opened 2016-02-03
Failing Behat tests if it generates errors in log.

grasmash:


grasmash:
This should fail because "Scenario: Filtering content by moderation state" generates the following warning the watchdog log: `Undefined index: #title` in `lightning_form_views_exposed_form_alter()`. 

I've used deprecated calls to `db_truncate()` and `db_select()` here because I frankly don't know how to inject the DB service into the Behat FeatureContext class.

grasmash:
Hm. I believe this uncovers a deeper issue. No Drupal functions appear to be available for calling from the lightning.behat.inc file. This same code executes successfully on my local machine, leading me to believe its stems from Behat configuration.

I would guess that Behat is not actually bootstrapping Drupal directly due to placement of the composer.json in the tests subdirectory. It may also be because `drupal_root` is set to a relative path, which I had thought wasn't possible--maybe something changed upstream in the Behat extension.

balsama:
Hmmm. I'd love to learn more about your local environment, @grasmash . Can I glean most of that information from bolt?

danepowell:
I cleaned this up and it's working pretty great over in #125. We just need to clean up the actual errors that Lightning is throwing during tests. I think this issue can probably be closed.

balsama:
This is older, but #125 has a working PR. Closing as dup.

PR #111, opened 2016-02-22
Probo

dzink:


phenaproxima:
Looks cool, but what is the advantage of using Probo over (or in addition to) Travis?

dzink:
@phenaproxima my mistake - meant to make this PR for the zivtech branch! We're just testing things out over here.

PR #118, opened 2016-04-22
8.x clean tests

balsama:


tests/features/bootstrap/panels.behat.inc
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Places a block into a Panels layout.
+   *
+   * @param string $plugin_id
+   *   The block plugin ID.
+   * @param string $category
+   *   The block's category.
+   *
+   * @When I place the :plugin_id block from the :category category
```
balsama, line 86: While were here, does it make sense to be able to specify which region the block is placed in?

balsama, line 86: @phenaproxima ^^

phenaproxima, line 86: So far we haven't needed that, so I'm thinking we could just add that when we do...?

PR #119, opened 2016-04-26
Adding front end assets to composer.json.

grasmash:
Currently, simply using composer to install lightning will not build the front end dependencies. Drush make must be used, or `npm install` must be called from the lightning directory. This causes automated tests to fail.

This PR uses composer to manage front end dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,15 +23,14 @@
     ],
     "scripts": {
         "post-install-cmd": [
-            "PhantomInstaller\\Installer::installPhantomJS",
-            "./bin/phing"
-        ],
-        "post-update-cmd": [
-            "./bin/phing push"
+            "PhantomInstaller\\Installer::installPhantomJS"
         ],
         "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold"
     },
     "extra": {
+        "asset-installer-paths": {
+          "npm-asset-library": "libraries"
+        },
         "installer-paths": {
             "docroot/core": [
                 "drupal/core"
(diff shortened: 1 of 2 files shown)


grasmash:
A few notes:
- This PR relies on https://github.com/francoispluchino/composer-asset-plugin/pull/216 being merged upstream
- This notably slows down composer builds, at least on the first run. This is due to npm package information being downloaded. It is cached and subsequent builds are faster.
- We do not have a mechanism to ignore recursive dependencies, such as jquery, jqueryui, and backbone.
- I believe this allows us to remove code. Namely the bower.json file, possibly package.json, and a few Phing targets.

grasmash:
francoispluchino/composer-asset-plugin#216 has been merged upstream. @balsama Can you restart travis tests and see if we get a pass?

balsama:
@grasmash errored: https://travis-ci.org/acquia/lightning/builds/125854511#L512

grasmash:
One more time? :p

balsama:
@grasmash Same result: https://travis-ci.org/acquia/lightning/builds/125854511

grasmash:
Very strange, it's claiming that the underscore package being downloaded is corrupt. Download the same package locally for me is fine. Not sure what would cause that.

grasmash:
https://github.com/francoispluchino/composer-asset-plugin/issues/100

balsama:
hmmm. and PR #119 doesn't have a cache on travis - presumably because it never finished?

danepowell:
See #129 and #131

PR #125, opened 2016-06-14
Behat should catch PHP errors.

danepowell:
We've noticed some PHP errors (mostly warnings and notices) especially around node creation coming from Lightning modules. It would be great if automated tests could catch these errors before they ever show up in downstream projects.

I expect this first Travis test to fail due to the existing PHP notices and warnings needing to be cleaned up.


danepowell:
Ack, I didn't see that #103 is trying to accomplish the exact same thing. And both are failing for the same reason, that the db functions aren't available in the Behat context. I guess we should close one or the other of these issues.

danepowell:
Yeah, I give up for now. I just don't know enough about contexts in D8 to debug this. All I can say for sure is that this works fine on Bolt-based D8 projects, but doesn't work for Lightning. Which makes me think it's something specific to Lightning's behat.yml. The only conspicuous difference I see is that Lightning disables autoload, whereas Bolt (and projects based on it) do not.

balsama:
Can we make this non-destructive? Right now, it wipes out the watchdog table which isn't obvious or expected - so it could lead to problems.

As for the errors, @phenaproxima indicated that drupal isn't bootstrapped during `@BeforeSuite` which seems to make sense.

danepowell:
The truncate is just there because Drupal tends to throw a lot of errors on installation that aren't necessarily related to Lightning. We can see what happens without it.

I'll try to get around the bootstrap issue, I might need some guidance though.

danepowell:
This is catching errors now, but it's catching too many errors... including some generated during installation and not related to Lightning. The next step would be to grab the current timestamp during a beforescenario and then only query for errors following that timestamp.

danepowell:
It seems like this is working as intended now, and properly catching PHP errors that Lightning is throwing during tests. Can someone from the Lightning team please assist with cleaning up those errors?

danepowell:
Looks like one of the errors has been fixed, there's still two left.

danepowell:
The causes of the errors are https://www.drupal.org/node/2664682 and https://www.drupal.org/node/2786493

danepowell:
Seems like EntityEmbed was fixed but now the Lightning team needs to update something on their end. And the panelizer patch is still an open issue.

danepowell:
I revised this to be non-destructive. It works very well... almost too well, in fact! It's catching a lot of the existing errors in Lightning, which is to be expected.

Some of these are not trivial to fix, so I've removed those erroring tests.

@balsama would you be comfortable merging this as-is and opening separate issues to restore the failing tests? Otherwise we'll likely never be able to merge this. The errors are never going to go away on their own. Alternatively, we might be able to add a tag to skip error checking on "known bad" tests, but that almost defeats the point.

Also, I'm not sure how to get this to pass on the previous Lightning versions, or why previous versions are being tested at all as part of pull requests...

PR #126, opened 2016-06-17
updating drupal-scaffold to 2.0.0 to help windows users

atannerd:
drupal-scaffold uses rsync in previous versions which causes issues for windows users.  By upgrading to 2.0.0 windows users do not need rsync

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -128,7 +128,7 @@
         "drupal/inline_entity_form": "^8.1.0",
         "drupal/metatag": "^8.1.0",
         "drupal/token": "^8.1.0",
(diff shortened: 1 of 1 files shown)


balsama:
Cool. Only problem is that users of Lightning Project will get conflicts since it requires ^1.2. (We can update that simultaneously, but existing users will still get conflicts until they edit their requirements since Lightning Project is used as a template.)

Any idea why this is major release of drupal-scaffold?

balsama:
https://www.drupal.org/node/2751541#comment-11308697

PR #129, opened 2016-06-21
Adding front end assets to composer.json.

danepowell:
This is identical to #119, which should have been good except for the failing Travis test. I suspect this was a transient issue, a corrupt tarball got cached in Travis or something, I'm hoping starting fresh will fix it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -102,4 +102,3 @@
         "npm-asset/jquery-appear-poetic": "0.3.6"
     }
 }
-
--- a/composer.json
+++ b/composer.json
@@ -23,15 +23,14 @@
     ],
     "scripts": {
         "post-install-cmd": [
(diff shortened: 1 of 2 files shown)


danepowell:
It seems Travis isn't running at all now. No idea why.

Additionally, we will need to merge 8.x-1.x and run `composer update` to resolve merge conflicts. But when I do this, composer throws an error:

> strpos() expects parameter 1 to be string, array given

grasmash:
Yeah, we need to pull from dev-master: `"fxp/composer-asset-plugin": "dev-master#b2cc94c8018a7f4d61d1e0ab0d211a2713daa4fe"`

To make sure we have the commit from https://github.com/francoispluchino/composer-asset-plugin/pull/216

grasmash:
Trying here #131

PR #130, opened 2016-06-21
Composer install should fail on patch failure.

danepowell:
This should fail Travis tests until https://www.drupal.org/node/2752375 is resolved.

Once that's resolved and merged into this PR, tests should pass, and this will prevent this from recurring.


danepowell:
It looks like Travis is running patch 2.6.1. The binary format problem that caused 2752375 only affects patch version 2.7.5 and later. So Travis won't catch this particular type of failure, but it could still catch many other types of patch failure.

balsama:
Merged. Thanks!

PR #131, opened 2016-06-21
Building front end assets with composer.

grasmash:


grasmash:
Hey hey! It worked!

danepowell:
I think `composer.lock` needs to be updated

danepowell:
Matt points out that even if composer.lock were updated, this upstream issue would be a blocker:
https://github.com/francoispluchino/composer-asset-plugin/issues/7

composer.json
```diff
@@ -137,6 +140,11 @@
         "drupal/entity_browser": "8.1.0-alpha3",
         "drupal/views_infinite_scroll": "8.1.0",
         "drupal/media_entity_document": "^8.1.0",
-        "drupal/video_embed_field": "^8.1.0"
+        "drupal/video_embed_field": "^8.1.0",
+        "fxp/composer-asset-plugin": "dev-master#b2cc94c8018a7f4d61d1e0ab0d211a2713daa4fe",
```
balsama: Composer doesn't let you pull specific commit hashes for dependencies of dependencies. So when lightning is a dependency of BLT or Lightning Project, this won't work.

Can you pull 1.1.4 and patch whatever functionality is missing?

PR #132, opened 2016-06-21
Ensure that composer.lock is kept up to date.

danepowell:
It's easy to make a change to `composer.json` and forget to update `composer.lock` as well, as with #131. This simple test should catch such problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ before_script:
   - cd ..
 
 script:
+  - composer validate --no-check-all --ansi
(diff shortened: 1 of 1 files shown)


phenaproxima:
I like this idea. Seems like a small, easy win for us. @balsama, what do you think? @danepowell, can you merge upstream changes?

danepowell:
I don't have merge access on the lightning repo.

balsama:
Committed. thanks.

PR #134, opened 2016-07-01
8.x upgrade path ci

balsama:


balsama:
@phenaproxima 
Nit: The phing target and references to it should be `update`, not `upgrade`.

PR #136, opened 2016-07-25
Removes views_infinite_scroll patch. 2701881

kylebrowning:
Patch was committed, so it no longer applies on fresh projects.

https://www.drupal.org/node/2701881


kylebrowning:
#138  Is better.

PR #138, opened 2016-07-25
Update views_infinite_scroll to 8.x-1.2

naveenvalecha:
Refer to issue-137

I have not tested the views_infinite_scroll functionality yet as my lightining setup is about to complete.


kylebrowning:
WE don't ave to remove the composer.json patches as well?

bhasselbeck:
@naveenvalecha this has to update composer.json as well.  We should open another pull request for this.

bhasselbeck:
@kylebrowning @naveenvalecha here is a PR for updating composer.json https://github.com/acquia/lightning/pull/139/files Not sure if I need to also generate a lock file, or if a new one will be generated for the change.

naveenvalecha:
@kylebrowning Thanks Sure. will update the PR

kylebrowning:
@naveenvalecha Yup, but lightning is including this patch specifically, https://github.com/acquia/lightning/blob/8.x-1.x/composer.json#L103

Which is causing this issue, this patch needs to be removed and composer.lock updated as well.

Unless I'm mistaken.

naveenvalecha:
updated the PR. Thanks @kylebrowning 
cc/ @balsama

composer.json
```diff
@@ -141,7 +137,7 @@
         "drupal-composer/drupal-scaffold": "^2.0.0",
         "drupal/pathauto": "^8.1.0",
         "drupal/entity_browser": "8.1.0-alpha6",
-        "drupal/views_infinite_scroll": "8.1.0",
+        "drupal/views_infinite_scroll": "8.1.2",
```
phenaproxima: Can this be ^8.1.0? We don't use pinned versions unless we're applying a patch, and since this PR thankfully removes a patch, we no longer need to pin the version. :)

naveenvalecha: Sure @phenaproxima Updated the PR!

phenaproxima:
Excellent! @balsama OKed this, so I've merged it and credited @naveenvalecha on drupal.org. Thanks!

gollyg:
I think that this issue highlights a potential problem with using non-specific releases against patched modules. This situation would be avoided if the requirement pointed to the specific point release. 

Perhaps we need a policy that when applying patches to a module - the module reference must be to a specific point release, ensuring that we do not get breakages when patches are incorporated into referenced modules?

balsama:
Lightning does pin patched modules to specific releases. See our [Dependency Version Constraints](http://lightning.acquia.com/blog/dependency-version-contraints) flow chart.

[Lightning 1.02 requires version 8.1.0 of Views Infinite Scroll](https://github.com/acquia/lightning/blob/04b967e8b8fa6d65710faeb4ca2b8d3914b947cf/composer.json#L144) because it is patched.

PR #140, opened 2016-07-29
Fix for undefined weight issue breaking functional tests.

sylus:
See: https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Entity!entity.api.php/function/hook_entity_extra_field_info/8.2.x
#141


balsama:
That failure looks like a fluke. Rerunning and I'm happy to merge if it passes.

Issue #135, opened 2016-07-22
Panels bandaid patch breaking save after previewing blocks with Panelizer


mortona42:
Steps to reproduce:
1. Add new Landing page
2. Add content via IPE
3. Click save > Save as custom

The save icon keeps spinning and never saves the page.

mortenson:
Don't mean to hijack the issue - but I've debugged this for a bit and have narrowed the issue down:

Easiest replication steps:

1. Use IPE on any Panelized Node
1. Click "Manage Content" > "Forms" > "Search Form"
1. Click "Toggle Preview", then "Add"
1. Save as Custom or Default
1. See 500 error in the console

The minimum environment I could get to replicate this was:

1. Install the dev release of Panels and Panelizer, enable Panelizer, and Panelize a Content Type
1. Apply https://www.drupal.org/files/issues/bandaid.patch to Panels and rebuild cache
1. Follow replication steps above

Further debugging let me see that at the point of IPE saving, the Panels Display Variant ID was different than what was being edited, which means the bandaid code is probably receiving the wrong context from somewhere.

mortenson:
I'm thinking of ways to fix this so that the Panelizer maintainers would actually commit it (rather than using bandaid-2.patch or something) - Looking back at https://www.drupal.org/node/2701433 - would a more permanent fix be using private tempstore in IPE?

The only reason IPE shares tempstore between users is that I had ambition about having Google docs-style editing between multiple users, but there's no plans to actually add that.

It's worth noting that if we moved to private tempstore, the Lightning test at https://github.com/acquia/lightning/blob/8.x-1.x/tests/features/layout/panelizer.feature#L42 would still fail, as if two Nodes are using the same default and you start editing one, there's no way for Panelizer to really know that you're intended on saving it as a custom display. This seems like expected behavior to me, and wouldn't be too jarring for users already on the bandaid. It sounds like the main issue was that one user's work leaking into another user's work, which private tempstore should fix.

Thoughts?

balsama:
Thanks for reviving this, Sam.

We wrote that test you linked to after a specific complaint on a PS project (cc @danepowell). I think it's a fairly common scenario (especially for projects that have a big "data entry" phase) and that we'd likely get more issues reported if we reintroduced it.

It sounds like that test would only fail because it's the same user though. Is that correct? If that's the case, then I imagine we could rewrite the test and users could avoid the problem by using best practice (i.e. not using the same account).

mortenson:
[...]

PR #145, opened 2016-08-05
Removed the band-aid patch for Panels in favor of using private tempstore

mortenson:
See #135


mortenson:
Important note from the issue, as this does change behavior for existing Lightning users:

> Users who won't like this change are ones that are used to editing multiple unsaved Landing Pages (that were previously using the default) at once, but with Workbench Moderation they can just put a Landing Page into draft, make a change, save as custom, and go about their day.

mortenson:
The test I modified is working - but the wizard tests are failing now. Not sure how we want to re-roll those, so I'm going to leave the PR open for Lightning team input.

Edit: Basically need to know if I should patch the Panelizer Wizard patch to use PrivateTempStore, or update the Drupal.org issue with an entirely new patch (which would delay that issue until https://www.drupal.org/node/2779813 is closed).

balsama:
@mortenson I don't think we have a preference either way. This issue hasn't come up outside of this single report, so we're happy to wait for d.o#2779813 if that would be a more proper fix.

I'm happy to help reroll the tests too if the only reason they're failing is minor step changes. (not sure if that's the case or not)

mortenson:
I tried patching the Panelizer default patch but it didn't pass tests - I'll need to roll back to the minimum commit (https://github.com/acquia/lightning/pull/145/commits/f07a820eac502e1712d1f5e07d1b144696353db9) and debug the errors manually.

balsama:
Is there a chance these tests are failing due to slight changes in step definitions.

@mortenson you're more familiar with the exact changes. If you think that it's just minor changes (e.g. clicking here vs clicking there) I'm happy to manually test and adjust the tests accordingly.

mortenson:
The only thing I'm intending to change is what TempStoreFactory Panels IPE (or the backend editors) use, so no test definitions need to change as the HTML Behat is interacting with should be unchanged.

I'm starting to think that because of the number of patches and dependencies involved with editing Panelizer defaults, I may have to think about this a bit more and implement a different solution. Even if PrivateTempStore was used and the tests pass, existing users would see newer, possibly weirder, behavior when using Panels IPE.

Tomorrow I'll draft an issue to discuss how we want to remove the bandaid, instead of plowing forward with this specific fix. We need to determine what behavior users expect when using Panels IPE + Panelizer + multiple defaults + multiple editors. There are a lot of edge cases to consider.

That said, I still think Panels IPE should use PrivateTempStore, so I'll keep the Drupal.org issue open regardless of what we decide on.

mortenson:
Closing this for now as debugging the test failures and patch dependencies may be a bigger scope than this PR intended. The Panels issue is in (https://www.drupal.org/node/2779813), and we can carry the long-term band-aid removal discussion in the issue (#135).

PR #152, opened 2016-09-22
Lightning Media Image: Display id updated to image_browser for Image …

purushotamrai:
…Browser Media View.

Issue Details:
Line No. 163 : lightening_media.install file creates **new display with id entity_browser** for Media View  
Line No. 60 : lightening_media_image.install file creates new display with **same id entity_browser** for Media View

Also, Image Browser config included in Lightening Media Image Feature points to image_browser display of media view while the above approach gives 'entity_browser_2' as display id to that display.


ajitdev (commented):
This looks okay to me 👍

phenaproxima:
I closed #152 because in looking at it, I'm not 100% sure that there is actually a problem here.

PR #158, opened 2016-10-07
Switching PhantomJS to Selenium.

grasmash:


phenaproxima:
@grasmash, I think this needs another reroll...

.travis.yml
```diff
@@ -1,4 +1,7 @@
+sudo: true
```
phenaproxima, line 1: This is supposed to be `sudo: required` according to Travis docs.

.travis.yml
```diff
@@ -40,8 +53,9 @@ install:
   - export COMPOSER_EXIT_ON_PATCH_FAILURE=1
 
   # MySQL Options
-  - mysql -e 'SET GLOBAL wait_timeout = 5400;'
-  - mysql -e "SHOW VARIABLES LIKE 'wait_timeout'"
+  - mysql -u root -e 'CREATE DATABASE drupal;'
+  - mysql -u root -e 'SET GLOBAL wait_timeout = 5400;'
+  - mysql -u root -e "SHOW VARIABLES LIKE 'wait_timeout'"
```
phenaproxima, line 58: I wonder if we can remove this line. @balsama, what do you think?

.travis.yml
```diff
@@ -40,8 +53,9 @@ install:
   - export COMPOSER_EXIT_ON_PATCH_FAILURE=1
 
   # MySQL Options
-  - mysql -e 'SET GLOBAL wait_timeout = 5400;'
-  - mysql -e "SHOW VARIABLES LIKE 'wait_timeout'"
+  - mysql -u root -e 'CREATE DATABASE drupal;'
```
phenaproxima, line 56: Can this line be merged with the next line?

.travis.yml
```diff
@@ -61,12 +75,13 @@ install:
...
   - drush runserver --default-server=builtin 8080 &>/dev/null &
-  - phantomjs --webdriver=4444 > /dev/null &
+  - mkdir tmp
+  - selenium-server-standalone -port 4444 -log tmp/selenium2.log &
   - sleep 5
 
 script:
   - cd ./profiles/lightning
-  - behat --strict
+  - behat --strict --stop-on-failure
```
phenaproxima, line 84: I didn't know --stop-on-failure was a thing, but I'm really glad you told me.

.travis.yml
```diff
@@ -40,8 +53,9 @@ install:
   - export COMPOSER_EXIT_ON_PATCH_FAILURE=1
```
phenaproxima, line 53: @balsama, has this ever actually done anything?

phenaproxima:
I think 8.x-1.x needs to be merged back into this PR, but other than that I think it looks good.

.travis.yml
balsama: yeah - that's almost definitely left over from whatever this was based on originally. yank it.

.travis.yml
```diff
@@ -19,17 +22,27 @@ env:
...
+  # Initialize xvfb (see https://docs.travis-ci.com/user/gui-and-headless-browsers)
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+  - sleep 3
+  - LATEST_CHROMEDRIVER=$(wget -q -O - http://chromedriver.storage.googleapis.com/LATEST_RELEASE)
+  - wget http://chromedriver.storage.googleapis.com/$LATEST_CHROMEDRIVER/chromedriver_linux64.zip
+  - unzip chromedriver_linux64.zip
+  - chmod +x chromedriver
+  - mkdir -p $HOME/.composer/vendor/bin
+  - mv -f chromedriver $HOME/.composer/vendor/bin/
```
phenaproxima, line 45: Do we need to start chromedriver or anything, or does it just need to be in the PATH?

.travis.yml
```diff
@@ -19,17 +22,27 @@ env:
...
   - composer --version
+  # Initialize xvfb (see https://docs.travis-ci.com/user/gui-and-headless-browsers)
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+  - sleep 3
+  - LATEST_CHROMEDRIVER=$(wget -q -O - http://chromedriver.storage.googleapis.com/LATEST_RELEASE)
+  - wget http://chromedriver.storage.googleapis.com/$LATEST_CHROMEDRIVER/chromedriver_linux64.zip
+  - unzip chromedriver_linux64.zip
+  - chmod +x chromedriver
+  - mkdir -p $HOME/.composer/vendor/bin
```
phenaproxima, line 44: Can chromedriver_linux64.zip be deleted?

phenaproxima (approved):
I have merged this into the 8.x-selenium branch of Lightning and made changes there. Once it passes Travis, I think this is good to go into 8.x-1.x!

.travis.yml
balsama: It's just a safety net. There was a scenario where we were patching a module but we didn't lock it to a specific version. So the module updated and the patch failed to apply. The tests still passed in that case because the new version of the module included the commit that was in our patch. This will alert us if that happens again.

phenaproxima:
This is being merged with the 8.x-selenium branch. Thank you, @grasmash!

PR #178, opened 2016-10-13
Support prefixing for manual URL aliases when Pathauto is enabled

phenaproxima:
This PR fixes two URL alias prefixing scenarios when Pathauto is enabled:
1. If there are no applicable Pathauto rules and you enter a URL.
2. There _are_ applicable Pathauto rules, but you decide to bypass Pathauto and enter an alias manually.

This PR makes prefixing work as expected under both of these circumstances.


phenaproxima:
Merged in ba636f40c9351932fe3a12206b13ea40620af2e3.

PR #179, opened 2016-10-14
Fix file copy error when Multiversion is enabled on a new install

phenaproxima:
This issue is described over at https://www.drupal.org/node/2782239.

In a nutshell: if you create a completely new Lightning instance and start uploading normal files into it (i.e., into the public files directory) -- which Lightning does whilst installing its media component -- and then install Multiversion, you'll get weird warnings like this one:

File migrate://YOUR_FILE could not be moved/copied because it does not exist.

I traced this to a bug in core: FileSystem::mkdir() doesn't properly handle trailing slashes, which causes the error. I've filed a patch: https://www.drupal.org/node/2818031

This PR brings that core patch into Lightning.


phenaproxima:
Merged in ba636f40c9351932fe3a12206b13ea40620af2e3.

PR #186, opened 2016-10-19
Move Lightning Core config from install to optional

naveenvalecha:
Move Lightning Core config from install to optional


phenaproxima:
This is now fixed as of bc28695.

PR #190, opened 2016-10-26
Place "Page title" block in Seven theme

TravisCarden:
The page title is missing from the admin theme. I have copied the relevant YAML file right out of the core `standard` install profile.


phenaproxima:
We had a snafu in the main branch due to Chromedriver having a new version released recently. That is now fixed. Please pull upstream changes, then Travis _should_ pass this PR.

phenaproxima:
I think this is OK to go. Thank you, @TravisCarden!

PR #191, opened 2016-10-26
Make node authors replicate across workspaces

phenaproxima:
Fixes #173.

This was due to a bug in Replication. It is now fixed in HEAD, with test coverage.

The relevant d.o issue in Lightning: https://www.drupal.org/node/2817231
The relevant d.o issue in Replication: https://www.drupal.org/node/2820105

PR #192, opened 2016-10-26
Merge PR #2392 from Drush

phenaproxima:
According to #191, Drush's make-convert command is currently broken when dealing with certain Composer branch aliases, resulting in busted-ass makefiles being generated by our packaging script. The problem in Drush was fixed by drush-ops/drush#2392, so we need to merge that fix into our own Drush as a patch. That is what this here PR does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,13 @@ env:
 addons:
   apt:
     packages:
-    # Trusty does not include MySQL by defualt.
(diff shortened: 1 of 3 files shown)


balsama:
Given this passes travis, I think this is ready to merge.

phenaproxima:
Travis is hurling random failures here and there, but they are not related to this. It's just Travis being insane. Merging.

PR #199, opened 2016-11-01
Denote locked moderation states when editing a workspace

phenaproxima:
Lightning has the concept of locked workspaces, but when you edit a workspace, there is no way of knowing which of the available moderation states are considered locked. This PR adds a little lock icon (a standard Unicode character, it turns out) to any locked moderation state when they are listed on the edit form of a workspace.


balsama:
I really like this. Locked workspace states get confusing.

When I first saw it, I thought maybe the lock meant that I could transition to that workspace though. Let's add a little legend below that describes what the locked icon means. RTBC from me once we add that.

Thanks! This is another nice small win.

phenaproxima:
This had random failures on Travis (what else is new), but it seems to work well under manual testing (there is no automated test coverage because it's not a functional change) and considering that most builds passed just fine, I am merging it with @balsama's blessing.

PR #200, opened 2016-11-01
Remove Status column from workspace list

### LINKED ISSUE #184: Remove the Status column from the list of workspaces
It doesn't add any real value, and it confuses users. What does "active" mean? Its connotation is dangerously close to "published".

There should be one canonical place to check what the current workspace is, and that is the administrative toolbar. Lightning Preview already takes over the workspace list controller, so it should ice the Status column.

phenaproxima:
Low-hanging fruit -- fixes #184.


balsama:
This looks great. Nice small win. RTBC from me as soon as travis passes.

PR #201, opened 2016-11-03
Use database snapshots for testing update paths

phenaproxima:
Up until now, our approach to testing Lightning's update paths has been to install an old version of Lightning (by checking out a tag), then re-installing the latest HEAD on top of that. It works, but it's been quite messy and @grasmash believes it is the cause of our endless random failures on Travis. I'm still not certain that's what's at fault, but what *is* clear is that our update path testing is messy as hell.

@grasmash came up with the brilliant idea of simply using database snapshots to test the update paths. Upon each new release, we can install Lightning cleanly, dump the database to SQL and compress it, then store the archive in our repo. Then during CI we can simply import that database and boom -- we've got old version of Lightning upon which to test the update path. Nice and painless. No need to install dependencies twice. No need for overly-complex Phing targets. It's a win.


build.xml
```diff
@@ -163,23 +126,6 @@
     <delete file="libraries" />
   </target>
 
-  <!-- Builds a Lightning code base from legacy Drush make files. -->
```
balsama, line 166: Why are we getting rid of the build-legacy target? This still has the same relevance it did when we created it.

modules/lightning_features/lightning_media/lightning_media.module
```diff
@@ -22,9 +22,18 @@ function lightning_media_form_entity_browser_media_browser_form_alter(array &$fo
...
  */
 function lightning_media_entity_type_alter(array &$entity_types) {
-  // Files should never be workspace-aware. Besides, they ALWAYS conflict during
-  // replication, for reasons as yet unknown.
-  $entity_types['file']->set('multiversion', FALSE);
+  if (\Drupal::moduleHandler()->moduleExists('multiversion')) {
+    // Files should never be workspace-aware. Besides, they ALWAYS conflict during
+    // replication, for reasons as yet unknown. But because Multiversion is
+    // basically a one-way ticket, only opt file entities out of Multiversion if
+    // they haven't been migrated into Multiversion's storage yet. There really
```
balsama, line 29: Should this have it's own issue? I'm trying to understand when a file entity would not have yet been migrated to multiversion's storage. I know from anecdotal experience that files created after installing multiversion so still conflict, so I'm wondering if this would even address that problem.

.travis.yml
```diff
@@ -71,19 +71,25 @@ install:
...
   - composer global require phing/phing
-  - phing build -Dversion=$VERSION -Dbranch=$TRAVIS_COMMIT
-
-  # Install Lightning.
-  - phing install -Ddb.database=drupal -Dversion=$VERSION
-  # Execute database updates if we're not building from HEAD.
-  - if [ $VERSION != "HEAD" ]; then phing update; fi
+  # Build the Lightning code base.
+  - composer install
+  # Install Lightning cleanly so that settings.php will be created properly.
```
balsama, line 78: It would be nice if there was a way to just generate the setting.php file and not install drupal here. Then if no corresponding DB is found below, do the full install.

.travis.yml
```diff
@@ -71,19 +71,25 @@ install:
...
+  - phing install -Ddb.database=drupal
+  - cd docroot
+  # If there is a database fixture for the version being tested, recreate the
+  # database from it. We need to destroy the existing database in order to
+  # prevent table collisions.
+  - FIXTURE=$TRAVIS_BUILD_DIR/tests/fixtures/$VERSION.sql.bz2
```
balsama: I'd be nice if we had a standard way of generating these databases. Maybe just a phing target that does a clean install, enables lightning_preview and does a drush sql-dump that generates a file with the proper name and ext. I see this simple step getting screwed up at least once if we don't.

balsama (changes requested):
Overall, this looks great. I like the idea.

Nothing here is a blocker, but I'd like to see your thoughts on the notes. thanks.

build.xml
phenaproxima: True. I got rid of it because I've never used this target for anything, ever (if you have, I can restore these lines), so I decided it was just cluttering up the build file. The only thing that builds Lightning using drush make is drupal.org and our package target has worked beautifully for that.

modules/lightning_features/lightning_media/lightning_media.module
phenaproxima: We can't remove it. Unfortunately, unconditionally opting files out of Multiversion unconditionally breaks the update path, so this cannot be merged without this change. If someone has already installed Preview before we opted files out of it, they're going to have to live with file conflicts -- although, truth be told, that's a problem that should be fixed in Multiversion. But hey, experimental, right? :)

.travis.yml
phenaproxima: I had the same thought and looked for something like that, but unfortunately there doesn't seem to be a Drush command (or flag to site-install) that will do it. It's not as simple, it turns out, as just putting a database in settings.php; the installer also generates hash_salt and possibly a few other things, which means the path of least trouble here is just to use drush site-install. It's annoying, but still a lot cleaner than it was before.

.travis.yml
phenaproxima: Yeah, I thought about that as well. It's certainly worth a shot, and I don't mind trying to write such a target, but one thing to be aware of is that it will need to blow away any existing Lightning DB in order to work properly.

build.xml
balsama: Yeah - we still have a need for it. Let's add it back in.

modules/lightning_features/lightning_media/lightning_media.module
balsama: That makes sense. Thanks.

.travis.yml
balsama: There's [this: Settings.php generator](https://www.drupal.org/project/settingsphp), but I think our time is better spent other places than getting that stable :)

.travis.yml
balsama: Yeah I think this is worth it. It would be something we ran just after a new release before the subsequent release. Your caveat is noted and I think expected.

phenaproxima: Okay. I added a new target, called memorize, which will generate the snapshot. It requires the code base to be built (i.e., git checkout followed by composer install) before it is run -- it can't do that stuff on its own because if it did a git checkout, build.xml would be overwritten with an old version that hasn't got the memorize target :)

The procedure for generating these, then, is basically this:

```
$ git checkout 8.x-1.14
$ composer install
$ phing memorize -Dversion=1.14
```

This should create tests/fixtures/1.14.sql.bz2 from a clean install.

I'm also adding a target to import an existing snapshot (again, this requires an already-built code base):

```
$ phing remember -Dversion=1.14
```

build.xml
phenaproxima: Done.

.travis.yml
```diff
@@ -71,29 +71,31 @@ install:
...
   # Start Selenium and dump its ginormous log into a temporary file.
-  - mkdir tmp
-  - selenium-server-standalone -port 4444 -log tmp/selenium2.log &
+  - mkdir ../tmp
+  - selenium-server-standalone -port 4444 -log ../tmp/selenium2.log &
   - sleep 5
+  - cd ./profiles/lightning
 
 script:
   - composer validate --no-check-all --ansi
```
balsama, line 98: 53afe66553a6968e5194dbb4a07fc7e6db802745 changed this so now it's verifying that the lock file in docroot/profiles/lightning is up to date with that composer.json. Correct me if I'm wrong, but we want to verify that the lock file at repo root is up to date.

balsama (changes requested):
Just one nit about the verify lock file which was committed yesterday. Seems like that got changed in this branch.

.travis.yml
phenaproxima: They're the same, because phing push will copy the root composer.json and composer.lock into the profile directory.

balsama: Ah, right. We're basically testing a copy of a manipulated file for something. I guess you could argue the same for the entire codebase though :)

Fair enough.

balsama:
Two followup issues:
1. This failure bothers me:
   https://travis-ci.org/acquia/lightning/jobs/174181422#L2106

Can we open an issue to rewrite that without javascript? Or figure out another way to make it more reliable?
1. Open a followup issue to update our developer documentation (http://lightning.acquia.com/lightning-developer-guide). That should also get submitted to Lynette so it lives within the rest of the documentation too.

balsama (approved):

phenaproxima:
Yeah, that "Access denied" fail drives me crazy. I will look into rewriting it.

And I will open a follow-up to mention this stuff in the dev documentation once Travis passes everything and I merge this PR.

Issue #195, opened 2016-10-26
View modes should support help text


phenaproxima:
This is inspired by #194.

If we implement that, I think we should also implement the concept of view mode
descriptions. I think it would be helpful to authors if site builders could attach help
text to a view mode, explaining how the view mode is used and where, and perhaps what it
looks like, or anything else pertinent or peculiar. We could then display that help text
as the description of each select list presented by the Panelizer widget, as a way of
helping authors choose which layout they'd like to use. 

Additionally, any help text attached to a view mode should be displayed when editing that
view mode, via hook_help().

koleary:
That's a really good idea.

PR #202, opened 2016-11-04
Implement descriptions for display modes and user roles

phenaproxima:
Implements #195. Even has a test!

Since this feature is a general improvement and not layout-specific, it went into Lightning Core. Layout's integration with it will come after this is merged.


phenaproxima:
Implemented #214 in 8afafc7e0b84bbc6b5e1da66c4b3477fe40dd734.

modules/lightning_features/lightning_core/lightning_core.module
```diff
@@ -81,6 +110,50 @@ function lightning_core_disable_buttons(array $element) {
...
+function lightning_core_set_third_party_description(array &$form, FormStateInterface $form_state) {
+  $form_state
+    ->getFormObject()
+    ->getEntity()
+    ->setThirdPartySetting('lightning_core', 'description', $form_state->getValue('description'))
+    ->save();
+
+  // Help text is render cached, so clear the entire render cache right now so
+  // it will pick up the updated description. @see lightning_core_help()
+  // TODO: Try to use cache tags here so as to be less of a blunt instrument.
```
balsama: Let's go ahead and do that. Give the help block it's own cache tag and then just clear that. I could see some editor on prod making a tweak here and bringing the site to its knees.

balsama:
From #195:

> We could then display that help text as the description of each select list presented by the Panelizer widget, as a way of helping authors choose which layout they'd like to use.

Doesn't look like this was implemented. Any reason why not?

modules/lightning_features/lightning_core/lightning_core.module
```diff
@@ -81,6 +110,50 @@ function lightning_core_disable_buttons(array $element) {
...
+ */
+function lightning_core_form_alter(array &$form, FormStateInterface $form_state) {
+  $form_object = $form_state->getFormObject();
+
+  if ($form_object instanceof EntityFormInterface) {
+    $entity = $form_object->getEntity();
+
+    if ($entity instanceof EntityDisplayModeInterface || $entity instanceof RoleInterface) {
+      $form['description'] = [
+        '#type' => 'textarea',
```
balsama, line 128: The description looks kind of funny bumped up against the rest of the help text with no paragraph break. What if we changed this to a textfield instead of a text area and wrapped the output in a p tag when it's rendered with the help. A textfield is also more conducive to being used as a description for the select lists in panelizer default choices if and when we implement that.

FWIW, core vocabulary description fields are also textfields not textareas.

phenaproxima: I decided to leave it as a textarea, since it hard to say how verbose these explanations may need to be...but made it dramatically smaller and wrapped it in a P tag when it's displayed.

modules/lightning_features/lightning_core/lightning_core.module
phenaproxima: Done.

PR #203, opened 2016-11-04
Thumbnails for tweets

phenaproxima:
Finally! After much work and many more delays, this feature is finally here, completely integrated for the first time in Lightning. I don't think it needs any introduction.

It has no tests -- it's pretty hard to functionally test this, and testing it in Lightning would gain us very little since the feature is an amalgam of work that is slated for both Drupal core and Media Entity Twitter, both of which will provide effective test coverage for this. But it should work beautifully under manual testing in our dev build ;-)


phenaproxima:
@slashrsm and @koleary might find this interesting. :)

slashrsm:
Yay! I reviewed the patch: https://www.drupal.org/node/2650976#comment-11766123

balsama:
Nice. It'd be good to get that patch to Media Entity Twitter RTBC'd before merging this to minimize changes to that patch. Also, don't we have a D.O issue for this? Let's migrate this there.

phenaproxima:
Bringing in the RTBC patch (or maybe a completely new release of MET, if they're willing to roll one once it's committed) seems reasonable to me. @slashrsm's review didn't ask for any serious changes, so that should be doable.

audiojak:
YAY!

phenaproxima:
@slashrsm committed the patch: https://www.drupal.org/commitlog/commit/70995/e5d7c69861a23daafc244256e9c99e8afe1bb286

I'm not sure if a new release will be rolled, but if not, we'll need to pin the version. @slashrsm, what do you think?

phenaproxima:
New release tagged: https://git.io/vX0Tp

Let's not merge this PR as-is; I'll bring the latest Media Entity Twitter into it.

balsama:
Ok great. This is RTBC from me given that we integrate MET 1.3. Thanks.

phenaproxima:
Yup, already done, and in this PR to boot! Merged.

PR #204, opened 2016-11-06
zendframework/zend-stdlib in composer.lock pushes php dependency to 5.6

showards:
I wrote a little bit about this on <a href="https://www.drupal.org/node/2825170" title="zendframework/zend-stdlib in composer.lock pushes php dependency to 5.6 @ drupal.org">drupal.org</a> but this is my first time using github for this sort of collaboration, so please forgive me; I know not what I do.

That said, I was going to wait for Drupal's testbot to ensure zero regressions, but I didn't make any seriously substantive code changes.  The major changes were dependency updates.  If I'm not using github correctly, please let me know. :/


showards:
Oops, I forgot to write in the commit message drupal-composer/drupal-scaffold was also updated from 2.1.0 to 2.2.0. :/

phenaproxima:
Heh...this is actually a known issue. We discovered it a few weeks ago, but I'm not sure it is for us to address.

Why? Because Lightning itself has no direct dependency upon zend-stdlib; it is either one of core's dependencies, or it's a dependency of one of core's dependencies. Either way, due to its sloppy versioning, it pushed Lightning's minimum PHP version 5.6 when building with Composer. This does not affect the drupal.org tarball because that is built automatically using drush make behind the scenes, because drush make has no concept of dependencies or minimum PHP versions. It's just a dumb downloader.

We decided that this inconsistency was acceptable because PHP 5.5 is now well past its end-of-life and could (should) be considered dangerous/insecure. Personally, I think that Drupal core should have raised their own minimum PHP version to 5.6 to account for that, but they didn't because...reasons, I guess.

I'm not sure we'll be able to accept this PR as it is, because by changing the makefiles it would, in essence, forcibly regress Lightning version to 8.x-1.11, and that is not an option :) My feeling is that we should instead consider this a documentation issue -- we should officially declare Lightning's minimum PHP version to be 5.6 (which, again, I feel would be appropriate since we have effectively inherited the constraint from core, even if core itself claims, wrongly, that its minimum version is 5.5) and move on with our lives. But I do realize this is potentially contentious so I'd like to get @balsama's opinion about this.

showards:
I wasn't sure whether it was an intended or an unintended effect.  Please forgive the PR from github; I didn't realize 5.5 was already EOL status.  I am not too smart in that regard, I'll be the first to admit.

When I was testing all of this stuff, I noticed the build from drupal.org uses the appropriate version to escape any issues with regard to PHP versions.  I was just looking at ensuring my own build worked.  I wanted to start work on getting some sort of media_entity_audio.  (see <a href="https://www.drupal.org/node/2825075">#2825075</a>)  I just happened to run into these things while I started to work on those.

I was just trying to help those who may not have gotten there yet or those who may not have the knowledge to build newer PHPs than provided by the system; some systems are really old.  I certainly don't want to introduce regressions or even potential security holes by forcing versions susceptible to intrusion.  Oh, god, no.  I'm just starting to use the collaborative features of github and please forgive me if I fsck up from time to time.  (I just forked my first repo and committed to it in a public setting within the last 8 hours!  Yeah I know that's like, "BFD," but it's new to me.)

I will leave this open, but feel free to close it.  I'm not tied to it by any means; learning experience, if anything!

phenaproxima:
Please, don't apologize. I really appreciate anyone bringing any issues to our attention and any effort made by the community to improve Lightning!

balsama:
See https://www.drupal.org/node/2825326#comment-11766737

Issue #211, opened 2016-11-08
Authenticated users cannot see the administrative toolbar


phenaproxima:
I'm not entirely sure if this is hidden behind a permission or what, but by default authenticated users cannot even *see* the admin toolbar, even if they have permission to use some of its items. I think Lightning should grant this permission to all authenticated users by default.

phenaproxima:
Issue migrated to drupal.org: https://www.drupal.org/node/2825929

Issue #209, opened 2016-11-08
Authenticated role should be able to use administration theme


phenaproxima:
In the course of building a demo, I've had to switch in and out of a couple of user accounts with varying degrees of privilege. I cannot begin to describe how jarring it is that in certain circumstances I can see the (relatively nice and clean) Seven theme, and in others I can only use the main theme, which, more likely than not, lacks administrative styling and therefore looks like a hideous, barely-usable mess.

Lightning should change this, IMHO, and in new installs (not upgrades), it should give all authenticated users permission to use the administrative theme. Obviously that will not give them permission to use administrative pages, but if they DO need to use administrative pages, they should all be using the same theme.

phenaproxima:
@koleary, any thoughts?

koleary:
I totally agree. I have never been a fan of restricting authors admin experience via the theme. It's a blunt instrument instead of a scalpel. In addition to the problems Adam listed it's also condescending and makes authors feel like second-class citizens, not to mention creating a disconnect when admins have to explain something while simultaneously 'translating' instructions about what the author should see from the admin theme to the front end theme.

phenaproxima:
Issue migrated to drupal.org: https://www.drupal.org/node/2825932

Issue #208, opened 2016-11-08
Reviewer role cannot access content overview page


phenaproxima:
Well, this is freaking annoying. By default, the "reviewer" content role cannot access the /admin/content page, which means they effectively need to know the URLs of things to review. Super annoying oversight. Needs to be fixed in Lightning Core with a test.

phenaproxima:
Issue migrated to drupal.org: https://www.drupal.org/node/2825934

PR #213, opened 2016-11-08
Grant basic permissions to authenticated users

### LINKED ISSUE #212: Authenticated users cannot see moderation states
By default, authenticated users who have permission to log in and look at the list of content (which should be most users) will see an empty column for "Status" on /admin/content, no matter what the actual moderation state for that row is.

This is because authenticated users do not have the permission to view moderation states. Lightning should enable this permission for them out of the box. This can be easily fixed but it will need a test.

phenaproxima:
This will fix #212, #211, #209, and #208. Just needs tests.

### DIFF
--- a/modules/lightning_features/lightning_core/config/install/lightning_core.settings.yml
+++ b/modules/lightning_features/lightning_core/config/install/lightning_core.settings.yml
@@ -12,5 +12,6 @@ content_roles:
     enabled: true
     label: '? Reviewer'
     permissions:
+      - 'access content overview'
       - 'edit any ? content'
       - 'delete any ? content'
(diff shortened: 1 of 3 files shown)


phenaproxima:
It occurs to me that "authenticated user" could be a pretty damn broad swath of users, depending on the individual site; so maybe we should instead wrap these permissions into a new role -- not sure what to call it yet -- which grants these incredibly basic authoring permissions common to all users who need to actually modify the site.

rlnorthcutt:
+1

phenaproxima:
As of Lightning 2.0.4, this is no longer an issue.

PR #219, opened 2016-11-16
update dependencies

balsama:
List of lightning dependencies that are out of date in 1.12:

- entity_browser (alpha 10 -> 12)
- features (should auto update)
- media_entity_twitter (should auto update)
- metatag (should auto update)
- multiversion (alpha10 -> 12)
- view_infinite_scroll (should auto update)
- workspace (alpha3 -> 4)

All of the above seem to have been updated in other issues with the exception of entity_browser. So this updates that and a bunch of core dependencies.


phenaproxima (dismissed):
It looks like Entity Browser has tagged beta1, let's switch to that :)

balsama:
Oh good. I'll remove the constraint too then.

balsama:
Removed EB constraint and did some manual testing too.

Issue #194, opened 2016-10-26
Improve the author-facing Panelizer experience


phenaproxima:
While helping @audiojak build out a demo, I thought of a way to improve Panelizer's UX in Lightning.

Consider the Panelizer widget that gets presented on node forms for node types that have been panelized. If the Default view mode is allowing a choice of layouts, it just barfs out a pile of select lists, several of which are not relevant. Who on Earth wants to choose a layout for token rendering? Or search indexing (distinct from search result display)? Or RSS? My guess is nobody. Those extraneous select lists represent scary and confusing clutter, especially to non-technical authors.

Here's what I propose.

1) Lightning should introduce the concept of an "internal" view mode. Internal view modes are view modes used by the system for various things, but never rendered out to the user under normal circumstances. Tokens, search indexing, and RSS are the prime examples that ship with core. We would store a boolean "internal" flag on these view modes, in their third-party settings.

2) We would also completely disable Panelizer for any internal view mode. Internal view modes should not even expose the checkboxes that allow one to configure Panelizer. Marking a view mode internal after the fact would keep all of Panelizer's settings intact, but ~~disable Panelizer itself for that view mode~~ if the user ever disables Panelizer for that view mode, they will not be able to re-enable it.

3) When configuring an internal view mode, a message is displayed to tell the user that the view mode is internal and will never be seen by regular users.

I'd love to hear thoughts and opinions on it.

phenaproxima:
Blocked on #195.

balsama:
Just to be clear, the current solution is to never panelize the default display mode. I wonder if a much more simple solution would be to just prevent panelizing the default?

balsama:
Implemented in #223

PR #222, opened 2016-11-17
Add view mode descriptions to the Panelizer widget

phenaproxima:
Partially implements #194 with a test. Also of note is that this brings in the beginnings of the fabled Lightning Extension for Behat.


phenaproxima:
Issue created on drupal.org: https://www.drupal.org/node/2828638

balsama:
merged.

PR #224, opened 2016-11-18
Don't use git as download type in Drush make file

phenaproxima:
Fixes https://www.drupal.org/node/2827227.

I don't see any compelling reason to keep using git as the download type, if it is causing version strings to be excluded from info files -- which can actually cause real problems, as we're seeing in the case of Media Entity blocking database updates via update.php because it doesn't think that Entity API alpha3 or later is installed, when in fact it is.


Natshah:
Thank you :+1: 
Not showing the version was a real hassle for developers and the update managing.

composer.json
```diff
@@ -43,6 +43,9 @@
         "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold"
     },
     "extra": {
+        "packagist.drupal-composer.org": {
```
phenaproxima, line 46: This bit of magic was just merged into drupal-packagist by @webflo. It will cause the packagist to completely ignore our make file, thus freeing us to make it look and work however we want.

webflo:
👍

balsama:
Committed. Thanks @webflo for your responsiveness!

PR #225, opened 2016-11-21
Added error checking to behat tests.

danepowell:
Re-rolling #125


danepowell:
@balsama I finally have this working without having to modify existing tests at all. Could you please review this?

Going forward, you can ignore tests with known errors by adding the `@errors` tag

balsama:
This is great. Thanks for seeing this through. I guess - ideally - we'd have an issue for each of `@error` tags on D.O - at least for the ones outside of lightning_preview. We can follow up with that though.

PR #226, opened 2016-11-22
Fixed severe Panelizer bug

danepowell:
As I mentioned in the [d.o issue](https://www.drupal.org/node/2664682#comment-11786436), if you disable Field UI on a Lightning-based site right now, it will put your db into a near-unrecoverable state, because that Panelizer patch introduces a hidden dependency on Field UI.

The updated patch in this PR simply makes that dependency explicit, so no one else falls into that trap.

For anyone who's already encountered this bug and needs help, let me know and I can post recovery instructions.


phenaproxima:
Nice catch. +1 for this.

PR #227, opened 2016-11-25
Issue #2830050 by bkosborne: FileUpload widget should allow file exte…

bkosborne:
This is to resolve this issue: https://www.drupal.org/node/2830050

This modified the `BundleResolverInterface` to expose the `getPossibleBundles()` method. I also refactored things a bit to make better use of the `SourceFieldTrait`. Not sure how comfortable you are with the API change. It seems like it would have little effect unless other users of Lightning decided to create their own bundle resolvers w/o extending the abstract one this module provides.

As for testing, I'm unable to get any of the behat tests that require JS to run on my Mac unfortunately (I installed JDK, selenium standalone server, and the chromium driver, but I still get failure to connect errors when a javascript test tries to run =/)

### DIFF
--- a/modules/lightning_features/lightning_media/src/BundleResolverBase.php
+++ b/modules/lightning_features/lightning_media/src/BundleResolverBase.php
@@ -13,19 +13,14 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  */
 abstract class BundleResolverBase extends PluginBase implements BundleResolverInterface, ContainerFactoryPluginInterface {
 
+  use SourceFieldTrait;
+
   /**
    * The media bundle entity storage handler.
    *
    * @var \Drupal\Core\Entity\EntityStorageInterface
    */
   protected $bundleStorage;
-
-  /**
-   * The configurable field entity storage handler.
-   *
-   * @var \Drupal\Core\Entity\EntityStorageInterface
-   */
-  protected $fieldStorage;
 
   /**
    * BundleResolverBase constructor.
@@ -58,35 +53,17 @@ abstract class BundleResolverBase extends PluginBase implements BundleResolverIn
   }
 
   /**
-   * Returns all possible bundles for the field type(s) this plugin supports.
-   *
-   * @return MediaBundleInterface[]
-   *   Applicable media bundles, keyed by ID.
+   * {@inheritdoc}
    */
-  protected function getPossibleBundles() {
+  public function getPossibleBundles() {
     $plugin_definition = $this->getPluginDefinition();
 
     $filter = function (MediaBundleInterface $bundle) use ($plugin_definition) {
-      $field = $this->getSourceField($bundle);
+      $field = $this->getSourceFieldForBundle($bundle);
       return $field ? in_array($field->getType(), $plugin_definition['field_types']) : FALSE;
     };
 
     return array_filter($this->bundleStorage->loadMultiple(), $filter);
-  }
-
-  /**
-   * Returns the source field for a media bundle.
-   *
-   * @param \Drupal\media_entity\MediaBundleInterface $bundle
-   *   The media bundle entity.
-   *
-   * @return \Drupal\Core\Field\FieldConfigInterface
-   *   The configurable source field entity.
-   */
-  protected function getSourceField(MediaBundleInterface $bundle) {
-    $type_config = $bundle->getType()->getConfiguration();
-    $id = 'media.' . $bundle->id() . '.' . $type_config['source_field'];
-    return $this->fieldStorage->load($id);
   }
 
 }
--- a/modules/lightning_features/lightning_media/src/BundleResolverInterface.php
+++ b/modules/lightning_features/lightning_media/src/BundleResolverInterface.php
@@ -23,4 +23,12 @@ interface BundleResolverInterface {
    */
   public function getBundle($input);
 
+  /**
+   * Returns all possible bundles for the field type(s) this plugin supports.
+   *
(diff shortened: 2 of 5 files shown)


modules/lightning_features/lightning_media/src/BundleResolverBase.php
```diff
@@ -21,13 +23,6 @@
   protected $bundleStorage;
 
   /**
-   * The configurable field entity storage handler.
-   *
-   * @var \Drupal\Core\Entity\EntityStorageInterface
-   */
-  protected $fieldStorage;
-
```
bkosborne, line 29: This is provided by the SourceFieldTrait, no longer needed

modules/lightning_features/lightning_media/src/BundleResolverBase.php
```diff
@@ -58,35 +53,17 @@ public static function create(ContainerInterface $container, array $configuratio
...
-   *   Applicable media bundles, keyed by ID.
+   * {@inheritdoc}
    */
-  protected function getPossibleBundles() {
+  public function getPossibleBundles() {
     $plugin_definition = $this->getPluginDefinition();
 
     $filter = function (MediaBundleInterface $bundle) use ($plugin_definition) {
-      $field = $this->getSourceField($bundle);
+      $field = $this->getSourceFieldForBundle($bundle);
```
bkosborne, line 62: I added this new method to the `SourceFieldTrait` and removed the locally defined method `getSourceField` below

balsama:
This looks good to me. I'll give @phenaproxima a chance to look at it later in the week before merging since you said you have a workaround. Good catch.

modules/lightning_features/lightning_media/src/BundleResolverBase.php
```diff
@@ -13,6 +13,8 @@
  */
 abstract class BundleResolverBase extends PluginBase implements BundleResolverInterface, ContainerFactoryPluginInterface {
 
+  use SourceFieldTrait;
```
phenaproxima, line 16: I'm not sure about SourceFieldTrait being part of BundleResolverBase. It probably won't harm anything, but it isn't architecturally "pure", since the bundle resolver API shouldn't make assumptions about the media bundles in question.

modules/lightning_features/lightning_media/src/BundleResolverInterface.php
```diff
@@ -23,4 +23,12 @@
    */
   public function getBundle($input);
 
+  /**
+   * Returns all possible bundles for the field type(s) this plugin supports.
+   *
+   * @return MediaBundleInterface[]
+   *   Applicable media bundles, keyed by ID.
+   */
+  public function getPossibleBundles();
```
phenaproxima, line 32: Can we have a base implementation of this method in BundleResolverBase which simply returns all bundles, unconditionally?

modules/lightning_features/lightning_media/src/BundleResolverBase.php
bkosborne: Wasn't it essentially already part of `BundleResolverBase` though? It had a method `getSourceField` which I removed in favor of using the trait since it was mostly code duplication.

modules/lightning_features/lightning_media/src/BundleResolverInterface.php
bkosborne: `BundleResolverBase` already has an implementation of this which I mostly kept untouched. Instead of returning all possible bundles it limits the bundles to those that have a matching source field definition.

modules/lightning_features/lightning_media/src/BundleResolverBase.php
phenaproxima: Good point. That's what I'm seeing too. Not sure how it got so hashed up.

modules/lightning_features/lightning_media/src/SourceFieldTrait.php
```diff
@@ -32,4 +33,23 @@ protected function getSourceField(MediaInterface $entity) {
...
+   *
+   * @param \Drupal\media_entity\MediaBundleInterface $bundle
+   *   The media bundle entity.
+   *
+   * @return \Drupal\Core\Field\FieldConfigInterface
+   *   The configurable source field entity.
+   */
+  protected function getSourceFieldForBundle(MediaBundleInterface $bundle) {
+    $type_config = $bundle->getType()->getConfiguration();
+    if (!isset($type_config['source_field'])) {
```
phenaproxima: Let's change this to if (empty()) so we can account for any falsy value.

modules/lightning_features/lightning_media/src/SourceFieldTrait.php
```diff
@@ -32,4 +33,23 @@ protected function getSourceField(MediaInterface $entity) {
...
+   * Returns the source field for a media bundle.
+   *
+   * @param \Drupal\media_entity\MediaBundleInterface $bundle
+   *   The media bundle entity.
+   *
+   * @return \Drupal\Core\Field\FieldConfigInterface
+   *   The configurable source field entity.
+   */
+  protected function getSourceFieldForBundle(MediaBundleInterface $bundle) {
+    $type_config = $bundle->getType()->getConfiguration();
```
phenaproxima, line 43: To re-use this code, let's change getSourceField() to:

```
return $this->getSourceFieldForBundle($entity->bundle->entity)
```

phenaproxima:
This is looking great. I have a couple of fairly minor requests, then I think this can be merged.

Issue #231, opened 2016-11-29
New Site installs in Acquia Cloud Site Factory fail to use lightning.extend.yml if file is placed in sites/default.


msherron:
.. but the extend file will work fine if it's placed in sites/all. We should update the documentation to reflect that.

balsama:
I think we had an issue on D.O where it was explained that this is works as designed. Whether we could add further documentation remains undecided, but either way, the extender is being removed in the next release of lightning in favor of sub-profiles. See https://www.drupal.org/node/2855793#comment-11972290

PR #232, opened 2016-11-29
Adding note for ACSF users in lightning.extend.yml documentation (#231).

msherron:


balsama:
Hmm:

> For Acquia Cloud Site Factory users, this file must live in sites/all for your overrides to take effect

This isn't entirely accurate. The file needs to live in your site path _or_ the sites/all directory (and this is already documented within the lightning.extend.yml file itself). The fact that `/sites/default` is _not_ the site path for sites on ACSF is inconsequential.

I don't know what the site path is for sites on ACSF (probably `/sites/{SITENAME}/`). But whatever it is, [Extender.php looks there first before moving on to `sites/all`](https://github.com/acquia/lightning/blob/8.x-2.x/src/Extender.php#L42).

I think we can close this as "works as designed" but I'm open to feedback.

msherron:
ACSF doesn't use the normal `sites/{SITENAME}` multisite convention at all. It uses `sites/g` (The 'g' is a holdover from Gardens, I believe) for *all sites* which contains code to dynamically determine which site instance in the Factory is being requested. If you'd like, we can work out a way where you can see an example from my current project for reference.

How I got here: I had erroneously thought from reading the documentation that I could put lightning.extend.yml in `/sites/default`, since that seemed the most applicable at the time. This worked great for ACE (we'd been deploying to both but are standardizing around ACSF), local builds, and Travis - but when building a new site instance in ACSF, my overrides in the extend file weren't being used. 

You and another PS coworker mentioned sticking it in`sites/all` which works in all cases. It seems to me that we either want to make this more explicit in the documentation, *or* we could alter Lightning to be aware of `/sites/g`, but the first option seemed simpler.

I would agree with you that the `sites/default` issue is inconsequential except that ACSF (and our own BLT tool) still uses `sites/default` for adding settings overrides, further muddying the waters. Not sure why we haven't collapsed that yet, but I wouldn't be surprised if that changes in the future. Either way, I think we can be more explicit here to save future confusion.

balsama:
Closed pending https://www.drupal.org/node/2855793

Issue #205, opened 2016-11-07
Decouple content types from base install modules


brantwynn:
Providing basic page and landing page by default is definitely a good thing that this distro should do.

However, it limits people that wish to enable lightning without any content types installed.

Right now, lightning_core provides the Basic Page and lightning_layout provides the Landing Page.

I propose that two new modules are created.  lightning_landing and lightning_page -- both of which can be optional by turning them off via lightning.extend.yml

balsama:
This makes a lot of sense. Something to consider.

If we use the same pattern as we do for the lightning_media submodules, there will be no way to disable them via lightning.extend. See: https://github.com/acquia/lightning/blob/8.x-1.x/src/Form/ExtensionSelectForm.php#L236

Scope creep on lightning extend is something we've been worried about from day one - especially because it's an interim solution. But at the same time, I don't want to get into a list of 20 modules to pick and choose from when installing from the UI or in the lightning.extend list. So we'll probably need an additional key that accepts lightning submodules that should _not_ be enabled.

phenaproxima:
5330f88 was merged into 8.x-2.x; we need to backport to 8.x-1.x.

balsama:
Backported: https://github.com/acquia/lightning/commit/bc286951437a9c4d27395aea81e239316fa793e8

PR #237, opened 2016-12-05
Decouple the bundled content types

phenaproxima:
At the request of many people, this PR unbundles the page and landing page content types from their parent modules. They are now subcomponents (like all of Lightning Media's extensions), installed by default, but disable-able via the extender.


phenaproxima:
This addresses #205.

Issue #150, opened 2016-09-20
Make it possible to create display modes inline


phenaproxima:
The idea of a "display mode" being separate from a "view display" is, in my view, a total UX WTF and a classic example of Drupal's engineer-oriented UIs turning off new users. I think Lightning can do better.

I propose that we completely hide the "Display modes" pages, and instead add a text field to any entity type or bundle's Field UI "Manage Display" tab. This text field (actually a machine_name field) would be tucked away in the "Custom Settings" fieldset, and there would be a submit button next to it. When the button is clicked, the following things happen:
- A new display mode for that entity type is created, with that label and machine name.
- Custom view settings would be applied for that display mode, creating a new view display.
- The user is redirected to the page where they can configure the new view display.

We could also do this for the "Manage form display" tab; the only difference being it would create a new form mode and form display, rather than a view mode and view display.

balsama:
Migrated to https://www.drupal.org/project/lightning/issues/2924516

PR #238, opened 2016-12-07
Allow inline view mode creation

phenaproxima:
Implements #150. It's a quick usability win.

This also introduces a new way of doing form alterations using OO techniques -- better encapsulation and testability :) As well as various other unrelated refactoring.


phenaproxima:
I no longer think this feature is necessarily very useful, so I'm closing this out.

PR #240, opened 2016-12-08
Provide an upgrade path to Lightning 2

phenaproxima:
People using Lightning 1.x will need a way to go to Lightning 2, which uses packages.drupal.org -- but they might not understand how to edit their composer.json so that everything continues to function. Lightning should provide something to help them out. That is what this does.

Thanks @grasmash for his help with converting the version constraints!


composer.json
```diff
@@ -40,12 +41,10 @@
             "./bin/phing push",
             "./bin/phing package"
         ],
+        "upgrade-build": "Acquia\\Lightning\\UpgradeBuild::execute",
         "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold"
     },
     "extra": {
-        "packagist.drupal-composer.org": {
```
phenaproxima, line 46: I just realized that these lines should actually be kept.

balsama:
Merged

PR #241, opened 2016-12-08
Hide the unused "Alternative text" field on image media widget forms

TravisCarden:
See https://www.drupal.org/node/2767213#comment-11454241.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,13 +103,6 @@ will not need to do this.
 
 ## Known Issues
 
-### Media
-
-* If you upload an image into an image field using the new image browser, you
-  can set the image's alt text at upload time, but that text will not be
-  replicated to the image field. This is due to a limitation of Entity Browser's
-  API.
-
 ### Preview
 
 * This functionality relies on Multiversion, which:
--- a/modules/lightning_features/lightning_media/lightning_media.module
+++ b/modules/lightning_features/lightning_media/lightning_media.module
@@ -92,6 +92,39 @@ function lightning_media_entity_type_alter(array &$entity_types) {
 }
 
 /**
+ * Implements hook_field_widget_form_alter().
+ */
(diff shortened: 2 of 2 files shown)


balsama:
Thanks for this, Travis. This patch also seems to hide the alt text field when uploading an image via the media library in ckeditor though. Which would prevent users from entering alt text for inline images.

create_basic_page___site-install

Can we add an additional check in the form_alter hook to make sure we're on the correct implementation of the form?

TravisCarden:
Ah! Thank you, @balsama. I'll look into it as soon as I have a minute.

balsama (changes requested):
Already noted in the conversation tab, but this patch also seems to hide the alt text field when uploading an image via the media library in ckeditor though. Which would prevent users from entering alt text for inline images.

We'll need to add an additional check in the form_alter hook to make sure we're on the correct implementation of the form or take a different approach.

phenaproxima:
I'm not sure I understand what's gained by suppressing the alt field here. To me, we should fix the underlying problem, which is that Entity Browser can't propagate arbitrary data into the calling window...

TravisCarden:
@phenaproxima fixing the underlying problem is certainly preferable. The approach in this PR was proposed as a stop gap measure to prevent a broken-feeling UX until the underlying problem *could be* fixed.

TravisCarden:
I've had a chance to look at this again, and I don't see that there's any way to distinguish between contexts in the form alter hook. Seems like we need to either find a better implementation for the stop gap or actually fix the underlying problem. I don't have bandwidth to undertake either right now, I'm afraid.

ovillarroel:
Any news about this well-known issue?

TravisCarden:
Old and obsolete.

PR #248, opened 2016-12-14
Fix upgrade script to account for git dependencies

bkosborne:
Not all packages in a project's composer.json will have a notification URL (like those coming from a git repo)

Without this, the upgrade script was bombing it out when it came upon a package we have that's coming from a git repo we maintain/


balsama:
Ah. We were sure that we'd miss some scenario. Thanks. I'll merge this into 8.x-1.x. If anyone else comes across this, they can pull dev to update to 2.x.

Thanks!

PR #257, opened 2017-01-04
Issue #2838896: Update workbench moderation to 8.x-1.2

naveenvalecha:


drupal-org.make
```diff
@@ -76,7 +76,6 @@ projects[views_infinite_scroll][type] = module
 projects[views_infinite_scroll][version] = 1.3
 projects[workbench_moderation][type] = module
 projects[workbench_moderation][patch][] = https://www.drupal.org/files/issues/2668006-2.patch
-projects[workbench_moderation][patch][] = https://www.drupal.org/files/issues/workbench_moderation-invalid_service_definition.patch
 projects[workbench_moderation][version] = 1.1
```
phenaproxima: Why wasn't the version changed in this make file? Running ```composer update``` should have done that (see the post-update-cmd script in our composer.json).

phenaproxima: Ah, I take it back -- it seems we were pegged to 1.1 because we were patching it (duh). Anyway -- as far as I can tell, this PR is only removing the patch, but is not actually updating WBM to 1.2. You'll need to change the constraint in composer.json as well, and run composer update to re-generate the make files.

composer.json
```diff
@@ -120,7 +120,7 @@
         "drupal/page_manager": "1.0.0-alpha24",
         "drupal/panelizer": "3.0.0-alpha2",
         "drupal/scheduled_updates": "1.0.0-alpha5",
-        "drupal/workbench_moderation": "1.1.0",
+        "drupal/workbench_moderation": "^1.0",
```
phenaproxima: Unfortunately this won't do, because we are still patching Workbench Moderation. When we're patching anything, we have to pin to a specific version as a matter of course. So this should be 1.2.0.

naveenvalecha:
Accommodated all changes as requested.

balsama:
Thanks @naveenvalecha - committed in a87a32f4920a119db2843be2daad948d73db3339

PR #259, opened 2017-01-12
Entity Block + search api

balsama:


.travis.yml
```diff
@@ -79,11 +79,10 @@ install:
   # Import the fixture, if it exists.
   - phing remember -Dversion=$VERSION -Ddb.database=drupal -Ddb.user=lightning -Ddb.password=lightning -Ddb.host=127.0.0.1
   - cd docroot
-  - drush cache-rebuild
```
balsama, line 82: This was just added a few weeks ago. I can concur that it's not needed, but any idea why it was added then?

modules/lightning_features/lightning_core/config/optional/views.view.content_index.yml
```diff
@@ -0,0 +1,276 @@
+langcode: en
```
balsama: Can we add a description to this view?

modules/lightning_features/lightning_core/src/Element.php
```diff
@@ -0,0 +1,102 @@
...
+    foreach (RenderElement::children($element) as $key) {
+      if (is_array($element[$key])) {
+        $element[$key] = call_user_func([static::class, __FUNCTION__], $element[$key]);
+      }
+    }
+    return $element;
+  }
+
+  /**
+   * Applies the Oxford comma to an array of strings.
```
balsama, line 56: plus 1000 :)

modules/lightning_features/lightning_core/config/optional/views.view.content_index.yml
balsama: Why is this in config/optional? What would happen if I enabled lightning_search at some random point after I enabled lightning_core?

modules/lightning_features/lightning_core/modules/lightning_search/css/entity-search.css
```diff
@@ -0,0 +1,3 @@
+.ui-autocomplete-loading[data-entity-search] {
```
balsama: core/modules/system/css/components/autocomplete-loading.module.css around line 16 has the same style defined. Why doesn't our form pick up on that style?

balsama: ~~Just tried to inspect the form to see if a class was different and I got a PluginException:~~


    Drupal\Component\Plugin\Exception\PluginException: Plugin (entity_view) instance class "Drupal\lightning_core\Plugin\Block\EntityViewBlock" does not exist. in Drupal\Component\Plugin\Factory\DefaultFactory::getPluginClass() (line 97 of /Users/adam.balsam/Sites/balsama-lightning/docroot/core/lib/Drupal/Component/Plugin/Factory/DefaultFactory.php).

balsama (commented):
I stopped after I got an exception while trying to actually use the block :)

    Drupal\Component\Plugin\Exception\PluginException: Plugin (entity_view) instance class "Drupal\lightning_core\Plugin\Block\EntityViewBlock" does not exist. in Drupal\Component\Plugin\Factory\DefaultFactory::getPluginClass() (line 97 of /Users/adam.balsam/Sites/balsama-lightning/docroot/core/lib/Drupal/Component/Plugin/Factory/DefaultFactory.php).

.travis.yml
phenaproxima: I think it was for safety. I needed to remove it because in some cases, without the updates having been run, cache clearing will fail.

modules/lightning_features/lightning_core/config/optional/views.view.content_index.yml
phenaproxima: Nice catch! We certainly can.

phenaproxima: This shouldn't even be in Lightning Core. Must be an artifact.

modules/lightning_features/lightning_core/modules/lightning_search/css/entity-search.css
phenaproxima: The core autocomplete library brings in that basic styling, plus a lot of dross that we can't use (I know because I tried). This was the path of least resistance. I don't think we can import assets libraries we don't "own"...

modules/lightning_features/lightning_core/lightning_core.module
```diff
@@ -135,24 +139,10 @@ function array_reorder(array &$values, array $keys) {
  * @return array
  *   The renderable element with all buttons (at all levels) disabled.
  *
- * @internal
+ * @deprecated
```
naveenvalecha: When this deprecated method will be removed? in lightning 8.x-3.x?

phenaproxima: Yes. Lightning 2.x follows semantic versioning, so I can't remove that function until 3.x (whenever THAT is).

naveenvalecha: okay

balsama:
Closing for now. Need to work out architecture and dependencies first.

PR #261, opened 2017-01-16
Add lightningscaffold plugin class with scaffold method to add dirs

balsama:


balsama:
@phenaproxima brought up the fact that we could run `mkdir -p` directly from the scripts section of composer, but I kind of like the idea of this being usable on non *NIX environments. I'm pretty sure the rest of our composer install is at least.

balsama:
@phenaproxima did you fix this with a simple `mkdir`? Should this PR be closed?

phenaproxima:
Yes. phing install now creates the PHPUnit directories, so we can ice this.

PR #264, opened 2017-01-20
Add Search API to Lightning

phenaproxima:
It's time.

This PR adds Search API to Lightning as a sub-component of Lightning Core, called Lightning Search. It creates a fairly standard index of rendered content, using the database backend (if available -- Search API DB is installed during a normal site install, but not during an update, and nothing in this PR has a hard dependency on it because most users will probably change to something more advanced, like Solr). Special sauce is added to support landing pages, which need a little help due to their unique nature.

There's also a lot of unrelated refactoring and cleanup -- moving useful functions into helper classes, removing redundant cruft, that sort of thing.


balsama:
4e8419fef5feb8ec743a2d12025813dfc4e79450

PR #272, opened 2017-02-07
Panelizer update

balsama:
Beta1 of panelizer adds a "save as default" perm which changes the "save as custom" link to just _save_ when you don't have permission to save any other way. So I had to make sure that tests that looked for that link also have the save a default permission.


tests/features/search.feature
```diff
@@ -15,9 +15,9 @@ Feature: Site search
     And I should not see "Zombie 2"
     And I should see "Zombie 3"
 
-  @javascript
+  @javascript @failing
```
phenaproxima: Should this failing tag be here?

tests/features/search.feature
```diff
@@ -15,9 +15,9 @@ Feature: Site search
...
   Scenario: Indexing and searching for landing pages
-    Given I am logged in as a user with the landing_page_creator,landing_page_reviewer roles
+    Given I am logged in as a user with the landing_page_creator,landing_page_reviewer,layout_manager roles
```
phenaproxima: Why do they get the layout_manager role as well?

tests/features/search.feature
balsama: Fixed

tests/features/search.feature
balsama: Without it, the save as custom button is just "Save" - and the step definition assumes the link is "Save as custom". Layout manager grants permission to "Save as default" too which changes the custom link back to "Save as custom.

phenaproxima (approved):

PR #273, opened 2017-02-15
Support profile inheritance and make it easier for Lightning to be a base profile

balsama:
Known issues:
* subprofil command requires drupal to be installed
* Need migration script


balsama:
Known issues:
* Entity block appears to be enabled but doesn't show up in the IPE for fresh installs (updates work fine)
* Drupalconsole subprofile command requires drupal to be installed to work (as does the generate profile command on which it was based)
* Need Lightning Extend -> subprofile migration script

lightning.info.yml
```diff
@@ -43,8 +43,21 @@ dependencies:
   - diff
   - metatag
   - search_api_db
-  # Our own core functionality.
+  # Lightning components
```
phenaproxima: Let's differentiate this from subcomponents by saying something like "Core Lightning components" or "Main Lightning components". And we should mention that Lightning Core is always required and that everything else depends on it.

lightning.info.yml
```diff
@@ -43,8 +43,21 @@ dependencies:
...
   - lightning_core
+  - lightning_layout
+  - lightning_media
+  - lightning_workflow
+  # Lightning subcomponents
```
phenaproxima: I think we should explain that these components depend on the main components, and mark which main components they depend on. Because otherwise someone might try to do something crazy like install Lightning Search without Lightning Core.

lightning.services.yml
```diff
@@ -9,3 +9,15 @@ services:
     class: '\Drupal\lightning\FormHelper'
     arguments:
       - '@element_info'
+
+  lightning.subprofile:
```
phenaproxima: Can we rename this to lightning.subprofile_command for clarity?

modules/lightning_features/lightning_preview/lightning_preview.info.yml
```diff
@@ -8,3 +8,4 @@ experimental: true
 dependencies:
   - lightning_workflow
   - workspace
+components: {}
```
phenaproxima: Not sure we need this if it's empty?

modules/lightning_features/lightning_workflow/lightning_workflow.info.yml
```diff
@@ -9,3 +9,4 @@ dependencies:
   - lightning_core
   - scheduled_updates
   - workbench_moderation
+components: {}
```
phenaproxima: Ditto.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+use Drupal\Core\Extension\InfoParser;
+use Drupal\lightning\Generator\SubProfileGenerator;
+use Drupal\lightning_core\Element;
+use GuzzleHttp\Client;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+
+class SubProfileCommand extends ProfileCommand {
```
phenaproxima: Needs a class doc comment for coding standards compliance.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+
+  use CommandTrait;
+  use ModuleTrait;
+  use FormTrait;
+  use ConfirmationTrait;
+
+  /**
+   * @var string \Drupal::root().
```
phenaproxima: Needs a description, and \Drupal::root() should not be in the type hint.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+   */
+  protected $appRoot;
+
+  /**
+   * @var array
```
phenaproxima: Needs a description, and ideally the type hint should say what is in the array, i.e. string[] -- assuming it's just an array of strings. If it's a multidimensional array of arrays or something like that, then 'array' is fine.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+   */
+  protected $topLevelComponents;
+
+  /**
+   * @var array
```
phenaproxima: Missing a description.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+      ->setName('lightning:subprofile')
+      ->setDescription($this->trans('Generate a subprofile of Lightning'))
+      ->setHelp($this->trans('The <info>lightning:subprofile</info> command helps you generate a new subprofile of Lightning'))
+      ->addOption(
+        'profile',
+        '',
+        InputOption::VALUE_REQUIRED,
+        $this->trans('The name of the subprofile')
+      )
+      ->addOption(
```
phenaproxima: Should this maybe be an argument, rather than an option? It seems to me that arguments should be used for required values, generally, and the machine name is about as required a value as we'll have in this command.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        'machine-name',
+        '',
+        InputOption::VALUE_REQUIRED,
+        $this->trans('The machine name of the subprofile (lowercase and underscore only)')
+      )
+      ->addOption(
+        'description',
+        '',
+        InputOption::VALUE_OPTIONAL,
+        $this->trans('Subprofile description')
```
phenaproxima: Nit: Maybe just "Description", since it's obvious that we're making a subprofile.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        'dependencies',
+        false,
+        InputOption::VALUE_OPTIONAL,
+        $this->trans('commands.generate.profile.options.dependencies')
+      )
+      ->addOption(
+        'excluded_dependencies',
+        false,
+        InputOption::VALUE_OPTIONAL,
+        $this->trans('Top-level components of Lightning to exclude separated by commas.')
```
phenaproxima: Can this be an abstracted string?

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        'profile',
+        '',
+        InputOption::VALUE_REQUIRED,
+        $this->trans('The name of the subprofile')
+      )
+      ->addOption(
+        'machine-name',
+        '',
+        InputOption::VALUE_REQUIRED,
+        $this->trans('The machine name of the subprofile (lowercase and underscore only)')
```
phenaproxima: Should be an abstracted string.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        'excluded_dependencies',
+        false,
+        InputOption::VALUE_OPTIONAL,
+        $this->trans('Top-level components of Lightning to exclude separated by commas.')
+      )
+      ->addOption(
+        'excluded_subcomponents',
+        false,
+        InputOption::VALUE_OPTIONAL,
+        $this->trans('Lightning sub-components to exclude separated by commas.')
```
phenaproxima: Prefer abstracted string.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+      )
+      ->addOption(
+        'distribution',
```
phenaproxima: As we discussed the other day, we should probably remove this option entirely. Anyone seeking to create a distribution can do that manually.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        '8.x'
+      )
+      ->addOption(
+        'dependencies',
+        false,
+        InputOption::VALUE_OPTIONAL,
+        $this->trans('commands.generate.profile.options.dependencies')
+      )
+      ->addOption(
+        'excluded_dependencies',
```
phenaproxima: Can this be excluded_components?

Also, if these will be command-line options, don't we want this to be excluded-components, not excluded_components? I think the command-line options should be used with dashes, like --excluded-components=foo,bar,baz -- not sure if Symfony handles that for us automagically, though?

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        $this->trans('The machine name of the subprofile (lowercase and underscore only)')
+      )
+      ->addOption(
+        'description',
+        '',
+        InputOption::VALUE_OPTIONAL,
+        $this->trans('Subprofile description')
+      )
+      ->addOption(
+        'core',
```
phenaproxima: We can remove this option because it will always be 8.x.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        $this->trans('commands.generate.profile.options.distribution'),
+        false
+      );
+  }
+
+  /**
+   * ProfileCommand constructor.
+   */
+  public function __construct() {
+    $translater = new TranslatorManager();
```
phenaproxima: Can we add a little more white space to this method and be sure all the various calls are grouped together logically? Just for readability and clarity.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+   */
+  protected function interact(InputInterface $input, OutputInterface $output) {
+    $io = new DrupalStyle($input, $output);
+    $validators = $this->validator;
+
+    try {
+      // A profile is technically also a module, so we can use the same
+      // validator to check the name.
+      $profile = $input->getOption('profile') ? $validators->validateModuleName($input->getOption('profile')) : null;
+    } catch (\Exception $error) {
```
phenaproxima: Nit: The catch should be on its own line. And it's better to catch a more specific exception, if available, than just \Exception.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+      $io->error($error->getMessage());
+
+      return;
+    }
+
+    if (!$profile) {
+      $profile = $io->ask(
+        $this->trans('The name of the subprofile'),
```
phenaproxima: Can this use an abstracted string?

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        '',
+        function ($profile) use ($validators) {
```
phenaproxima: Minor thing, but this can be changed to just [$validators, 'validateModuleName']. No need to wrap it in a closure.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+          return $validators->validateModuleName($profile);
+        }
+      );
+      $input->setOption('profile', $profile);
+    }
+
+    try {
+      $machine_name = $input->getOption('machine-name') ? $validators->validateModuleName($input->getOption('machine-name')) : null;
+    } catch (\Exception $error) {
```
phenaproxima: Catch on a new line, and a more specific exception if there is one.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+      $io->error($error->getMessage());
+
+      return;
+    }
+
+    if (!$machine_name) {
+      $machine_name = $io->ask(
+        $this->trans('commands.generate.profile.questions.machine-name'),
+        $this->stringConverter->createMachineName($profile),
+        function ($machine_name) use ($validators) {
```
phenaproxima: This can also just be [$validators, 'validateMachineName']

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+          return $validators->validateMachineName($machine_name);
+        }
+      );
+      $input->setOption('machine-name', $machine_name);
+    }
+
+    $description = $input->getOption('description');
+    if (!$description) {
+      $description = $io->ask($this->trans('Subprofile description'), '');
```
phenaproxima: Should use an abstracted string if possible. And for consistency can we remove the word "Subprofile"?

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+      $input->setOption('description', $description);
+    }
+
+    $dependencies = $input->getOption('dependencies');
+    if (!$dependencies) {
```
phenaproxima: We might want a more robust check here, since things like ' ' will be truthy. Do we need to run trim() on $dependencies or will Symfony do that for us?

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+      if ($io->confirm($this->trans('commands.generate.profile.questions.dependencies'), true)) {
+        $dependencies = $io->ask($this->trans('commands.generate.profile.options.dependencies'), '');
+      }
+      $input->setOption('dependencies', $dependencies);
+    }
+
+    $excluded_dependencies = $input->getOption('excluded_dependencies');
+    if (!$excluded_dependencies) {
```
phenaproxima: Same thing here.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+      if ($io->confirm($this->trans('Would you like to exclude any of Lightning\'s top-level components from the installation? (Experimental components are excluded automatically.)'), false)) {
+        $excludable_components = self::getTopLevelComponents(['lightning_core', 'lightning_preview']);
+        $question = new ChoiceQuestion('List the top-level components you would like to exclude, separated by commas.', $excludable_components);
```
phenaproxima: Should be an abstracted string.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        $question->setMultiselect(true);
+        $io->writeln('The following components can be excluded: ' . Element::oxford($question->getChoices()));
```
phenaproxima: Abstracted string.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        $excluded_dependencies = $io->askChoiceQuestion($question);
+      }
+      $input->setOption('excluded_dependencies', $excluded_dependencies);
+    }
+
+    $excluded_subcomponents = $input->getOption('excluded_subcomponents');
+    if (!$excluded_subcomponents) {
+      if ($io->confirm($this->trans('Would you like to exclude any of Lightning\'s sub-components?'), false)) {
```
phenaproxima: Abstracted string.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        // component isn't already excluded.
+        $excludable_subcomponents = [];
+        $top_level_components = self::getTopLevelComponents();
+        $non_excluded_tlcs = array_diff($top_level_components, $excluded_dependencies);
+        foreach ($non_excluded_tlcs as $non_excluded_tlc) {
+          $tlc_subcomponents = self::getSubComponents($non_excluded_tlc);
+          $excludable_subcomponents = array_merge($excludable_subcomponents, $tlc_subcomponents);
+        }
+
+        $question = new ChoiceQuestion('List the sub-components you would like to exclude, separated by commas.', $excludable_subcomponents);
```
phenaproxima: Abstracted string.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+        $question->setMultiselect(true);
+        $io->writeln('The following sub-components can be excluded: ' . Element::oxford($question->getChoices()));
```
phenaproxima: Ditto.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+   */
+  protected function execute(InputInterface $input, OutputInterface $output) {
+    $io = new DrupalStyle($input, $output);
+
+    if (!$this->confirmGeneration($io)) {
+      return;
+    }
+
+    $profile = $this->validator->validateModuleName($input->getOption('profile'));
+    $machine_name = $this->validator->validateMachineName($input->getOption('machine-name'));
```
phenaproxima: Didn't we already do this in the interact() stage?

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,363 @@
...
+    $description = $input->getOption('description');
+    $core = $input->getOption('core');
```
phenaproxima: Should be removed, let's just hard code this to 8.x.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+      $machine_name,
+      $profile_path,
+      $description,
+      $dependencies,
+      $this->excludedDependencies
+    );
+  }
+
+  /**
+   * @param array $excluded_dependencies
```
phenaproxima: The method is missing a description, and $excluded_dependencies can probably get a more specific type hint (like string[])

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+   *   Dependencies to exclude.
+   *
+   * Adds the provided component(s) to the excluded list and, if the provided
```
phenaproxima: This long description should be above the parameter list.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+   * component is a top-level component, all of its subcomponents too.
+   */
+  protected function buildExcludedDependenciesList($excluded_dependencies) {
```
phenaproxima: $excluded_dependencies should be type hinted.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+    $excluded_dependencies_list = [];
+    foreach ($excluded_dependencies as $excluded_dependency) {
+      if (array_key_exists($excluded_dependency, $this->getLightningComponents())) {
```
phenaproxima: Static methods should not be called with $this.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+        $excluded_dependencies_list[] = $excluded_dependency;
+        if (in_array($excluded_dependency, $this->getTopLevelComponents())) {
```
phenaproxima: Same problem here.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+          // If its a top-level-component, add its subcomponents too.
+          $subcomponents = $this->getSubComponents(trim($excluded_dependency));
+          foreach ($subcomponents as $subcomponent) {
```
phenaproxima: Why not just call array_merge instead of foreaching here?

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+            $excluded_dependencies_list[] = $subcomponent;
+          }
+        }
+      }
+    }
+    $this->excludedDependencies = array_merge($this->excludedDependencies, $excluded_dependencies_list);
+  }
+
+  public static function getLightningComponents() {
```
phenaproxima: Missing a doc comment.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+   *
+   * Filter an array by keys.
+   */
+  private static function array_filter_key(array $input, $callback) {
+    $keys = array_keys($input);
+    $filteredKeys = array_filter($keys, $callback);
+    if (empty($filteredKeys)) {
+      return [];
+    }
+    $input = array_intersect_key($input, array_flip($filteredKeys));
```
phenaproxima: Nit: Can we return directly from array_intersect_key() rather than overwrite $input, for clarity?

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+    $this->excludedDependencies = array_merge($this->excludedDependencies, $excluded_dependencies_list);
+  }
+
+  public static function getLightningComponents() {
+    $appRoot = \Drupal::root();
+    $extensions = new ExtensionDiscovery($appRoot);
+    $extensions = $extensions->scan('module');
+
+    $lightning_extensions = self::array_filter_key($extensions, function($key) {
+      return strpos($key, 'lightning_') === 0;
```
phenaproxima: Maybe not the most reliable filter...I think we should filter by the path instead. Any extension located inside Lightning's modules/lightning_features directory (which we can divine by doing a separate scan for profiles) is a Lightning extension; everything else is not, even if it's prefixed with lightning_.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+    });
+
+    $lightning_components = [];
+    foreach($lightning_extensions as $machine_name => $lightning_extension) {
```
phenaproxima: Nit: Need a space after 'foreach'.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+      $InfoParser = new InfoParser();
```
phenaproxima: Nit: Should be $info_parser or $infoParser.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+      if (isset($info['components'])) {
+        $lightning_components[$machine_name]['subcomponents'] =  $info['components'];
+      }
+    }
+
+    return $lightning_components;
+  }
+
+  /**
+   * @param $input
```
phenaproxima: Method description should be at the top, and all parameters are missing type hints and descriptions.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+   * @param $callback
+   * @return array|null
+   *
+   * Filter an array by keys.
+   */
+  private static function array_filter_key(array $input, $callback) {
```
phenaproxima: $callback should have the callable type hint. Also, the method should be protected because there is no apparent reason to use private.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+    if (empty($filteredKeys)) {
+      return [];
+    }
+    $input = array_intersect_key($input, array_flip($filteredKeys));
+
+    return $input;
+  }
+
+  /**
+   * @param array $excludedComponents
```
phenaproxima: Missing description for the method.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+   *   An array of top-level components to exclude.
+   * @return array of top-level Lightning components.
```
phenaproxima: There should be a blank line before the return annotation.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+   */
+  public static function getTopLevelComponents($excludedComponents = []) {
```
phenaproxima: $excludedComponents should be type hinted.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+    foreach (self::getLightningComponents() as $component => $attributes) {
+      if (isset($attributes['subcomponents'])) {
+        $topLevelComponents[] = $component;
+      }
+    }
+    return array_diff($topLevelComponents, $excludedComponents);
+  }
+
+  /**
+   * @param string $topLevelComponent
```
phenaproxima: Missing method description, parameter descriptions, and description of why the exception is thrown.

src/Command/SubProfileCommand.php
```diff
@@ -0,0 +1,351 @@
...
+   * @return array of subcomponents of the provided top-level component.
+   *
+   * @throws \Exception
```
phenaproxima: Preferable to throw a more specific exception. It seems like this should be a LogicException.

src/Generator/SubProfileGenerator.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Drupal\lightning\Generator;
+
+use Drupal\Console\Generator\ProfileGenerator;
+
+class SubProfileGenerator extends ProfileGenerator {
```
phenaproxima: Missing class doc comment.

templates/profile/info.yml.twig
```diff
@@ -0,0 +1,26 @@
+name: '{{ profile }}'
+type: 'profile'
+description: '{{ description }}'
+core: '8.x'
+distribution:
```
phenaproxima: Distribution stuff should be removed entirely.

templates/profile/info.yml.twig
```diff
@@ -0,0 +1,26 @@
+name: '{{ profile }}'
+type: 'profile'
+description: '{{ description }}'
```
phenaproxima: Description is optional, so this should be wrapped in an if check.

templates/profile/install.twig
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Implements hook_install().
+ *
+ * Perform actions to set up the site for this profile.
+ *
+ * @see system_install()
+ */
+function {{ machine_name }}_install() {
+
```
phenaproxima: Nit: The extra blank line is not needed.

templates/profile/install.twig
```diff
@@ -0,0 +1,19 @@
+<?php
+
+/**
+ * @file
+ * Install, update and uninstall hooks for the {{ profile }} Lightning
```
phenaproxima: I think this will read a bit funny in real-life use. Can it just say "for the {{ profile }} profile"?

templates/profile/install.twig
```diff
@@ -0,0 +1,19 @@
...
+ * subprofile.
+ */
+
+/**
+ * Implements hook_install().
+ *
+ * Perform actions to set up the site for this profile.
```
phenaproxima: Hooks need no further explanation beyond "Implements hook_foo()."

phenaproxima (dismissed):
Looking good, I just have minor complaints and nitpicks.

modules/lightning_features/lightning_preview/lightning_preview.info.yml
balsama: SubProfileCommand::getLightningComponents() uses the presents of the components key to determine if this is a top-level components. Seems more logical than adding an additional "is_top_leve: true" value or something. But I'm open to suggestions.

modules/lightning_features/lightning_workflow/lightning_workflow.info.yml
balsama: see above

src/Command/SubProfileCommand.php
balsama: Yeah. I switched this to filter by modules that live in [PATHTOLIGHTNING]/modules/lightning_features. Removed helper function  array_filter_key too.

templates/profile/info.yml.twig
balsama: I think it's safe to say that if someone is extending Lightning, they expect the extending profile to a distribution too.

My experience is that (with the core base profile patch) if a codebase has exactly one distribution, that distribution will always get installed, regardless of whether a profile exists that lists that distributions profile as a base profile. So I think this _needs_ to be there. (Without the core patch, I think you should get a TooManyDistributionsException.

Let's confirm/discuss further.

balsama: Correction, TooManyDistributionsException isn't even a thing. (I thought it was as part of the "Write install_profile value to configuration and only to settings.php if it is writeable [#2156401]). 

So I'm actually not sure what happens here. Certainly, the lines between profile and distribution are further blurred by this.

lightning.profile
```diff
@@ -2,101 +2,5 @@
 
 /**
  * @file
- * Defines the Lightning Profile install screen by modifying the install form.
+ * The Lightning profile.
```
phenaproxima, line 5: Do we still need the .profile? Doesn't look like there's going to be anything in it.

modules/lightning_features/lightning_preview/lightning_preview.info.yml
phenaproxima: SubProfileCommand is now using isset($info['components']) to check for the presence of sub-components, so things that don't have any should not define the components key at all or the command will get confused.

src/Composer/SubprofileMigrationCommand.php
```diff
@@ -0,0 +1,151 @@
...
+      // non-interactively. Otherwise, echo the command and tell the user run it
+      // themselves.
+      $package = $this->composer
+        ->getRepositoryManager()
+        ->getLocalRepository()
+        ->findPackage('drupal/console', '*');
+
+      if ($package) {
+        $command = $this->composer->getConfig()->get('bin-dir') . '/drupal ' . $command;
+        $this->io->write($command);
```
phenaproxima: Once things are confirmed working under manual testing we should probably use ProcessExecutor to call Console directly here. That's on me, though.

src/Generator/SubProfileGenerator.php
```diff
@@ -0,0 +1,77 @@
...
+namespace Drupal\lightning\Generator;
+
+use Drupal\Console\Core\Generator\Generator;
+
+class SubProfileGenerator extends Generator {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function generate(
```
phenaproxima: I think the arguments to generate should be on one line, and type hinted where possible.

src/Generator/SubProfileGenerator.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Drupal\lightning\Generator;
+
+use Drupal\Console\Core\Generator\Generator;
+
+class SubProfileGenerator extends Generator {
```
phenaproxima: The class will need a doc comment to pass PHPCS.

src/Generator/SubProfileGenerator.php
```diff
@@ -0,0 +1,77 @@
...
+    $description,
+    $dependencies,
+    $excluded_dependencies
+  ) {
+    $dir = $profile_path . '/' . $machine_name;
+
+    if (file_exists($dir)) {
+      if (!is_dir($dir)) {
+        throw new \RuntimeException(
+          sprintf(
```
phenaproxima: Can we call realpath() when we define $dir, rather than repeatedly calling it in sprintf()?

src/Generator/SubProfileGenerator.php
```diff
@@ -0,0 +1,77 @@
...
+            'Unable to generate the subprofile as the target directory "%s" is not writable.',
+            realpath($dir)
+          )
+        );
+      }
+    }
+
+    $parameters = [
+      'profile' => $profile,
```
phenaproxima: I think we should rename this to 'name'. 'profile' is a bit ambiguous.

phenaproxima (changes requested):
This looks GREAT! I'm running out of things to nitpick.

lightning.profile
balsama: Agree. We don't need it.

src/Generator/SubProfileGenerator.php
balsama: Agree. I got confused by this yesterday.

balsama:
I've addressed the last two items above (lightning.subprofile -> lightning.subprofile_command and remove the .profile file). I'll open a new PR for these items against the subprofile branch.

PR #276, opened 2017-02-17
Move the Behat tests into the relevant components

phenaproxima:
As advertised. Let's take advantage of Behat's flexible configuration to move our test scenarios into the relevant modules, as individual test suites. Any tests that involve more than one Lightning component will remain in Lightning itself.

As of this PR, to run only Media tests (for example), you would do this:

```
$ behat --suite=media
```

@grasmash and @arknoll, I would like your blessing on this first -- will this change break BLT? If you can confirm that it will be OK with this change, we'll merge it.


arknoll:
@phenaproxima BLT currently has lightning behat tests turned off completely. Therefore, nothing on BLT should break.

phenaproxima:
@arknoll OK, it's good to know that it won't break BLT...but the reason it won't break BLT is a bit disturbing. Why is BLT disabling Lightning tests entirely? Is it something we did?

grasmash:
@arknoll Lightning tests are disabled by default for BLT-generated projects, but we do still test Lightning integration in BLT's own CI process.

phenaproxima:
Ah, that's clearer. @grasmash, with that being the case, is there anything in this PR that will break BLT's CI?

behat.yml
```diff
@@ -27,3 +26,10 @@ default:
         login_form_selector: '#user-login-form'
 imports:
   - behat.local.yml
+  - modules/lightning_features/lightning_core/tests/behat.yml
```
grasmash, line 29: I'm pretty sure this will break BLT integration. This introduces a new requirement for executing Lightning tests: importing Lightning-specific configuration via Behat. This is not performed automatically for us by the Drupal Extension, and so will need to be done manually for any project wishing to execute Lightning tests.

phenaproxima: Is there anything we could do to prevent this from breaking BLT? Or is that something BLT would need to adjust around?

phenaproxima:
Closing this PR since I had to rebuild my fork. Will open a new PR once #299 is merged.

PR #280, opened 2017-02-24
Remove Search API DB from Lightning profile dependencies

phenaproxima:
As advertised. With this PR, Lightning Search will actively install Search API DB if it's available.

This addresses https://www.drupal.org/node/2855075.


balsama:
bc1825e

PR #282, opened 2017-03-02
Enable pipelines build and tests

balsama:
This PR adds a pipelines yaml file that instructs pipelines to build lightning and run our tests. At the end, the resulting codebase is deployed to the cloudlightning201 sub.

Concerns:
* Pipelines doesn't support java so we need to use phantom to run our test over there. This means that we're reliant on the `jakoch/phantomjs-installer` package again which is hosted on bitbucket and fails to download _a lot_.

Notes:

1. The main GH repo (acquia/lightning) needs to be associated with pipelines for builds to start on commit (currently only my fork is enabled)
2. Currently, I'm using a personal free sub for pipelines (cloudlightning201).
3. While pipelines does deploy the code to cloud, currently ~~there is no way~~ it is not easy to automatically run a command (i.e. `drush updb`) after deployment. So it would be wise to run `drush @[alias] si` after committing to 8.x-2.x
4. In writing this, I realized that I'm not sure what happens after a build on a feature branch. Does the feature branch code get deployed to the cloud sub given it passes? Overwriting the existing codebase? I would really only want that to happen when I committed to 8.x-2.x

Pinging @phenaproxima for feedback. @bjaspan it would be good to get clarification on note no. 4 above too.


balsama:
I've decided to close this PR for the following reasons:

* I don't want to rely on `jakoch/phantomjs-installer`. (See "Concerns" above)
* I don't want the codebase deployed to cloud unless it's on the default branch. (See note no. 4 above)

balsama:
@bjaspan clarified that pipelines creates a branch "per branch". So pipelines won't actually deploy a feature branch, it will just push it to the repo. We're free to leave the pipelines generated pipeline-build-8.x-2.x deployed. Or even just 8.x-2.x.

PR #284, opened 2017-03-06
Add a UI to Lightning Layout for configuring entity blocks

phenaproxima:
Fixes https://www.drupal.org/node/2851583.

Didn't add a test because this is so straightforward.


balsama:
@phenaproxima Can we clear the config object's cache when that form is submitted? It's a little confusing because changes made via that form aren't visible in landinga page -> entity block until the cache is rebuilt.

phenaproxima:
Done.

balsama:
committed in 96bacebf23209e6dabb4dada098553cc0310b387

Thanks!

PR #285, opened 2017-03-06
Adjust the verbiage when reverting a revision

phenaproxima:
Fixes https://www.drupal.org/node/2854057.

This takes a slightly advanced approach; if the target revision is a forward revision, this will say "switch" instead of "revert". Otherwise, it'll still say "revert". Didn't add tests since this is such a cosmetic change, but could do that if desired.


balsama:
From 2854057:

If you follow the steps in the original issue, the expected result is:
* Forward revision is set as default revision
* Default revision is unpublished
* Six total revisions remain (image in IS shows six)

Actual results
* Default revision remains unchanged
* New forward revision created that is a clone of the forward revision we wanted to make default

PR #287, opened 2017-03-07
Allow content reviewers to see moderation states

phenaproxima:
This PR addresses two issues: https://www.drupal.org/node/2825928 and https://www.drupal.org/node/2825934, which began life as #212 and #208, respectively.

After discussing with @balsama, we agreed to add the "view moderation states" permission to the content reviewer role(s) to avoid polluting the permission space of all authenticated users. @balsama felt that was too heavy-handed, and I agreed. So we're adding it to the content reviewer role since the most visible place where this bug manifests is on the content overview page, and reviewers are explicitly granted permission to access that page.

Regarding #208, all I did here was add a test confirming that reviewers can access the content overview page. The d.o issue will be closed (cannot reproduce) once this PR is merged.


phenaproxima:
Oh, and -- this fix required a manual update step. I have added documentation for that, and the equivalent code to Lightning Dev.

balsama:
This looks good. Just waiting for tests to pass

PR #288, opened 2017-03-08
Use the media browser for media reference fields

phenaproxima:
We already did this for image fields, now we should do it for entity reference fields that reference a media asset. Such fields should by default use our media browser (distinct from the image browser) for a more consistent UX.

This PR does exactly that and adds a very basic test.


balsama:
PIpelines (job id 175f14a1-a8a8-4704-8554-b45c97dbbac6) failed on:

**Scenario:**
Block placement on non-default displays are preserved after re-saving the entity
**Step:**
And I save the layout
**Error:**
Element with named "link Save" not found. (Behat\Mink\Exception\ElementNotFoundException)

That looks like a random failure and I've logged it over here: https://docs.google.com/spreadsheets/d/1Sb4KTZgVwPalLnsYk9pBblCjRsW2I1NBwsj0XuZgOtk/edit#gid=0

phenaproxima:
Closing due to my having destroyed and re-created my fork. Will open another PR with this feature.

PR #290, opened 2017-03-10
Use EntityDescriptionInterface to describe view modes and roles

phenaproxima:
A while ago, Lightning added support for attaching arbitrary descriptions to user roles and entity display modes. It turns out that core has a thing called EntityDescriptionInterface, which does precisely what we tried to do. In the name of dovetailing with core, this PR refactors our description code to use EntityDescriptionInterface.


balsama:
It looks like pipelines job (job id: b19a4d92-4a39-47b6-bcb2-aefeffb526c4) just killed the behat process for no apparent reason:

    /tmp/scriptaLouVW: line 24:  2325 Killed                  ../../../bin/behat --stop-on-failure

Not really sure what that means.

PR #291, opened 2017-03-10
Miscellaneous clean-up and refactoring

phenaproxima:
I pruned out some old, irrelevant code, cleaned up some existing stuff, and refactored a few form modifications...just gardening.


drupal-org.make
```diff
@@ -66,9 +66,7 @@ projects[replication][patch][] = https://www.drupal.org/files/issues/2814055-2.p
 projects[replication][patch][] = https://www.drupal.org/files/issues/2820105-10.patch
 projects[replication][version] = 1.0-alpha5
 projects[scheduled_updates][type] = module
-projects[scheduled_updates][patch][] = https://www.drupal.org/files/issues/schedule_updates-save_type-2674874-2.patch
```
balsama, line 69: This is technically out of scope for this issue.

modules/lightning_features/lightning_core/src/Access/AdministrativeRoleCheck.php
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Drupal\lightning_core\Access;
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Routing\Access\AccessInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Session\AccountInterface;
+use Symfony\Component\Routing\Route;
+
```
balsama, line 11: Class docblock

balsama: I also don't see any usages of this class. I like it, but why do we need it?

modules/lightning_features/lightning_media/modules/lightning_media_image/lightning_media_image.module
```diff
@@ -8,17 +8,15 @@
 use Drupal\Core\Entity\Display\EntityFormDisplayInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\field\Entity\FieldConfig;
+use Drupal\lightning_core\OverrideHelper as Override;
```
balsama, line 11: Why is this classname aliased?

modules/lightning_features/lightning_media/modules/lightning_media_twitter/lightning_media_twitter.module
```diff
@@ -6,23 +6,21 @@
  */
 
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\lightning_core\OverrideHelper as Override;
```
balsama, line 9: Ditto. Why alias?

modules/lightning_features/lightning_media/modules/lightning_media_video/lightning_media_video.module
```diff
@@ -7,23 +7,21 @@
 
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\media_entity\MediaInterface;
+use Drupal\lightning_core\OverrideHelper as Override;
```
balsama, line 10: Here too

modules/lightning_features/lightning_preview/lightning_preview.install
```diff
@@ -51,3 +51,10 @@ function lightning_preview_update_8001() {
   $display['display_options']['filters']['current_workspace'] = unserialize('a:15:{s:2:"id";s:17:"current_workspace";s:5:"table";s:15:"node_field_data";s:5:"field";s:17:"current_workspace";s:12:"relationship";s:4:"none";s:10:"group_type";s:5:"group";s:11:"admin_label";s:0:"";s:8:"operator";s:1:"=";s:5:"value";s:0:"";s:5:"group";i:1;s:7:"exposed";b:0;s:6:"expose";a:10:{s:11:"operator_id";s:0:"";s:5:"label";s:0:"";s:11:"description";s:0:"";s:12:"use_operator";b:0;s:8:"operator";s:0:"";s:10:"identifier";s:0:"";s:8:"required";b:0;s:8:"remember";b:0;s:8:"multiple";b:0;s:14:"remember_roles";a:1:{s:13:"authenticated";s:13:"authenticated";}}s:10:"is_grouped";b:0;s:10:"group_info";a:10:{s:5:"label";s:0:"";s:11:"description";s:0:"";s:10:"identifier";s:0:"";s:8:"optional";b:1;s:6:"widget";s:6:"select";s:8:"multiple";b:0;s:8:"remember";b:0;s:13:"default_group";s:3:"All";s:22:"default_group_multiple";a:0:{}s:11:"group_items";a:0:{}}s:11:"entity_type";s:4:"node";s:9:"plugin_id";s:17:"current_workspace";}');
   $view->save();
 }
+
+/**
+ * Clears the entity type definition cache.
+ */
+function lightning_preview_update_8002() {
```
balsama, line 58: Why do we need this?

balsama (commented):
This looks great. A few requests for clarification and possible nit identified below.

drupal-org.make
phenaproxima: True. I have no excuse, just thought I'd get it out of the way.

modules/lightning_features/lightning_core/src/Access/AdministrativeRoleCheck.php
phenaproxima: I don't know; I'm following the pattern established in core. I'm not implementing an interface, though, so presumably I could remove it.

modules/lightning_features/lightning_media/modules/lightning_media_image/lightning_media_image.module
phenaproxima: Better readability (more English-like). "Override::entityClass()" reads more like a sentence than "OverrideHelper::entityClass()".

modules/lightning_features/lightning_preview/lightning_preview.install
phenaproxima: We need these because when we override the entity forms and handlers, we're changing the entity type definitions and need to force those changes to be picked up. Normally this would happen during a cache clear caused by a database update, but if there were no database updates, the changes would not be picked up and things would fatal left and right. So this prevents that from happening.

modules/lightning_features/lightning_core/src/Access/AdministrativeRoleCheck.php
phenaproxima: Whoops, misread the original comment. This class implements the new _is_administrator access check, which is being used in a couple of routing files also updated in this PR.

PR #292, opened 2017-03-14
Always show media previews

phenaproxima:
This PR addresses https://www.drupal.org/node/2825935, targeted for Lightning 2.0.6.

It adds a dynamic preview to certain media types that need it (i.e., tweets, instagrams, and videos). The preview appears when adding a media item of these types, or when editing an existing one. The preview is considered part of the entity form and can be moved around in the "Manage form display" page for each of these media types. It also works with IEF, although the position of the preview can't be adjusted by the site builder (an apparent limitation of IEF).


balsama:
Committed in 08e8498283f1d1efc57f19965c627004a22b8815

PR #293, opened 2017-03-14
Fix PHP warning on workspace list page

phenaproxima:
Long has it been that we've seen PHP warnings related to rawurlencode() when we visit the workspace listing page. It was so annoying that we wrote it in as a known issue and had our tests suppress the warning on purpose.

In the time since that happened, some lovely person patched the issue in Workspace and it has been committed. This PR brings that patch into Lightning.


balsama:
My only comment here is that we use the `@errors` tag to suppress legitimate PHP errors from failing our tests. (see https://github.com/acquia/lightning/pull/225) In an ideal world we'd fix the errors we're suppressing. But that's not always possible.)

That's very different that what we're using the tag for here - where we are specifically testing to ensure an exception is thrown.

So, in one case, we're using the `@errors` tag to work around a problem in our codebase. In the other, we're using it to work around a problem with our test suite.

A couple of thoughts:
* This might expose a different (very out of scope) issue with the way we handle attempts to edit config in the non-default workspace. A simple attempt probably shouldn't result in a PHP error.
* Is the current system that fails when an error is thrown correct? I think so - but that backs up the previous bullet that we should either approach workspace/config editing differently or test that functionality differently.

So, I'm happy to merge this because it makes our tests more reliable and therefore useful. But wanted to get thoughts and post this for posterity. @phenaproxima and I can discuss in the AM.

phenaproxima:
Everything you're saying makes perfect sense, but we are largely constrained by what we can get away with in Drupal. The exception-throwing is a catch-all way for us to prevent *any* config from being edited in non-standard ways -- the blocks listing page being a perfect example. Those are the sort of things that users will do, and then freak out when their live site is affected. So I think throwing the exception was the right thing to do. We could implement the same thing in a less error-ish and invasive way (i.e., adding validation handlers to the forms), but we can't possibly identify all the forms that would need such treatment.

Personally, I think the smartest thing for us to do -- that would allow us to get rid of the exception, or at least stop treating it like an error condition -- is to implement our whole config-mirroring idea: clone the active config store, top to bottom, when a workspace is created. Then at least config in a workspace is a) editable and b) preserved, separate from the live workspace. We can figure out replication later.

PR #294, opened 2017-03-14
Automate updating of component version numbers

phenaproxima:
We have a lot of components and sub-components, and the number seems to grow with every release. They all need consistent version numbering in their info files. This is a pain to do manually, so let's have a script to do it. That's what this PR adds.

### DIFF
--- a/src/Composer/ReleaseVersion.php
+++ b/src/Composer/ReleaseVersion.php
@@ -2,12 +2,12 @@
 
 namespace Acquia\Lightning\Composer;
 
-use Acquia\Lightning\ComponentDiscovery;
 use Composer\Script\Event;
-use Drupal\Component\Serialization\Yaml;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Yaml\Yaml;
 
 /**
- * A script to update the version number in Lightning's component info files.
+ * Updates the version number in Lightning's component info files.
  */
 class ReleaseVersion {
 
@@ -20,28 +20,13 @@ class ReleaseVersion {
   public static function execute(Event $event) {
     $arguments = $event->getArguments();
 
-    $core_package = $event->getComposer()
-      ->getRepositoryManager()
-      ->getLocalRepository()
-      ->findPackage('drupal/core', '*');
+    $finder = (new Finder())->name('*.info.yml')->in('modules');
 
-    $install_path = $event->getComposer()
-      ->getInstallationManager()
-      ->getInstallPath($core_package);
-
-    $app_root = realpath($install_path . DIRECTORY_SEPARATOR . '..');
-
-    $discovery = new ComponentDiscovery($app_root);
-
-    foreach ($discovery->getAll() as $component) {
-      $path = $app_root . DIRECTORY_SEPARATOR . $component->getPathname();
-
-      $info = file_get_contents($path);
-      $info = Yaml::decode($info);
-
+    /** @var \Symfony\Component\Finder\SplFileInfo $file */
+    foreach ($finder as $file) {
+      $info = Yaml::parse($file->getContents());
       $info['version'] = reset($arguments);
-      $info = Yaml::encode($info);
-      file_put_contents($path, $info);
+      file_put_contents($file->getPathname(), Yaml::dump($info));
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/Composer/ReleaseVersion.php
```diff
@@ -0,0 +1,48 @@
...
+
+    $discovery = new ComponentDiscovery($app_root);
+
+    foreach ($discovery->getAll() as $component) {
+      $path = $app_root . DIRECTORY_SEPARATOR . $component->getPathname();
+
+      $info = file_get_contents($path);
+      $info = Yaml::decode($info);
+
+      $info['version'] = reset($arguments);
```
balsama: Wrap the value in single quotes for strict yaml parse.

balsama (changes requested):
Besides the single quote comment, this script behaves a little differently than other similar scripts that we use for automation (e.g. phing package). Specifically:
* it makes its changes to the running codebase, not the repo
* It requires the running codebase to be built

I'm not sure what our options are, especially for the second bullet, since changing it would mean we couldn't rely on _any_ external packages. Thoughts?

src/Composer/ReleaseVersion.php
phenaproxima: If I did that, it would be encoded as "'8.x-2.05'". Wrapping in quotes is up to the encoder, not me.

balsama: Currently, it's being output as:

    version: 8.x-2.04

I guess we could pass the quotes in the argument, but it'd be better if the script just output it in the same format we currently have.

phenaproxima: Even if we passed the quotes in the argument, the encoder would still escape them. I don't think there's anything we can do in this case. The encoder simply needs to be smart enough to add the quotes, which means we need to patch the serialization component that ships with core.

phenaproxima:
I could probably adjust it to modify the repo files, rather than the ones in the installed code base. That should take care of the first point you raised.

As for the second point, we may be limited by what ExtensionDiscovery does. It really expects to be working with a coherent Drupal code base; you can't just point it at any old directory and have it find all extensions (if that's how it worked, it would have been about a hundred times simpler).

An alternative approach would be to completely abandon ComponentDiscovery for this purpose and use the Symfony Finder component instead. It would easily be able to trawl through the repo's modules directory for anything ending in *.info.yml. Which, given the starting directory, would be guaranteed to be nothing but Lightning components. Would that be better?

src/Composer/ReleaseVersion.php
balsama: Hyphen is specifically listed as a character that should be escaped - and cloud has had catastrophic problems with yaml parsing before. So I guess we can just pass the quotes in with the argument. Need to confirm that serialization component doesn't strip those too though!

phenaproxima: How will passing the quotes in with the argument help? I must be missing something here...I'll do a bit of testing.

balsama:
> would that be better

I think so. But, honestly, this level of automation is far, far better than what we had. The MVP is "automate updating the info files". And we've met that. So let's not bikeshed on this point. Thanks. This is great.

src/Composer/ReleaseVersion.php
balsama: I was assuming the argument was treated as a string. At a minimum I would think we should be able to include _escaped_ quotes in the string. 

On mobile ATM so haven't tested.

phenaproxima:
Decoupling it from ExtensionDiscovery would allow me to remove the weird FileCacheFactory and eval() hacks, so I'm going to do it. Simple is beautiful.

src/Composer/ReleaseVersion.php
phenaproxima: I tried several permutations and it does not work. The Symfony YAML component is not 100% compliant with the specification (a known issue), and it simply does not quote the version number. Trying to pass the quotes on the command line causes the quotes to either be escaped, or duplicated as the parser tries to escape them for you.

I found a workaround, though. Don't quote the version number on the command line; instead, the script will surround it with << and >> characters. This forces the YAML dumper to quote the version number; the script then strips out the << and >> stuff from the dumped YAML string. Clever me. So this is solved.

balsama (approved):

balsama:
Clearly, these yaml parsers know yaml better than me. Now it's changing `'8.x'` to `8.x` and `"SLick media handling..."` to `'Slick media handling...'` Neither of which will be a problem. Thanks for entertaining :)

PR #296, opened 2017-03-15
Remove Lightning Preview from mainline Lightning release

phenaproxima:
It's getting unwieldy to keep Lightning Preview, with all the extra code and crazy hacking that it entails, in mainline Lightning. It's an experimental component, and not installed by default. @balsama had the brilliant idea -- with which I enthusiastically and wholeheartedly agree -- to split Lightning Preview out into a special "experimental" branch of Lightning, which can integrate more bleeding-edge modules from both core and contrib in a safe space that will not affect the stable 8.x-2.x branch.

Another benefit of an experimental branch is that we can get a head start on integrating experimental things that are slated to become stable, so that when these features stabilize, they can be brought into the next stable Lightning release at short notice.

This PR completely removes Lightning Preview, along with any other mention of experimental components or features, out of Lightning. Once this is merged, we'll spin up an 8.x-2.x-experimental branch, along with instructions for those cowboys and cowgirls who would like it play with it.


balsama:
This is targeted for Lightning 2.1.0, which will be the first release on the core 8.3.x branch.

phenaproxima:
Lightning 2.0.6 has been released, so I'm going to re-do this PR.

Issue #266, opened 2017-01-26
Unable to use media library on media reference fields


leonk:
When setting a media reference field, and using the entity browser.  I'm unable to select any entities from the "view" page.

Steps to replicate:
1) Install lightning_media_image and all it's dependencies
2) Add a entity reference field, select "Other" as the reference type, then select media.
3) Allow image media entities to be selected
4) In the entity form display, set the field to use the entity browser widget.
5) Edit or add an entity.  When you are on the "view" tab you are unable to select any entities.

I've had a look into what is going on, and it seems that the browser.js is not included on the page (from the lightning_media.module).
So I tried including it on the page.  This still didn't work, looks like a call the Drupal.behaviors isn't fired once the entity browser is loaded.

balsama:
I don't think I follow entirely. Which entity browser are you selecting? (Image browser or media browser?) And what's the end goal?

If you're just uploading images, can you just use the image reference field instead? Lightning will automatically select the image browser Entity Browser in that case.

leonk:
Okay so it works if I reference the image entity directly, and not the media entity.

My use case is I need to reference the media entity, as we have a few additional fields for information we need to capture on the image.
I'm using the image_browser, but have modified it to include an inline entity form upload.

balsama:
Ok. That makes sense. We might have an existing issue that this could fall under. Stay tuned.

phenaproxima:
@leonk: From a cursory reading of this issue, I think that you may be running into something we did intentionally.

The image_browser entity browser was intended to work with core image fields, which are just specialized versions of file fields. Meaning that they expect to be referencing a file entity directly. The image browser does some jiu-jitsu around this, so that you can select and create media entities, but when you actually complete the selection, it returns the underlying file entities to the field.

The aforementioned jiu-jitsu is configurable; if you edit the image_browser entity browser and look at the Upload widget, you'll see that there is a checkbox labeled "Return source file entity". This is checked by default. If you uncheck it, the image_browser will return the media entity rather than the file entity -- but this will instantly break the image browser break on any image field in which it is used.

I would instead suggest that you duplicate and modify the media_browser entity browser. It is configured to return media entities directly, and should therefore work with media reference fields.

phenaproxima:
#288 addressed this. I will re-open a new version of that PR shortly.

balsama:
#298 now.

phenaproxima:
I hope this helps :)

PR #298, opened 2017-03-16
Media reference fields should use the media browser by default

phenaproxima:
Re-constitutes #288, addressing #266.


balsama:
I haven't tested on a fresh install, but upgrading from HEAD of 8.x-2.x, the media browser is not automatically selected for media reference fields. Did a little debugging and the lightning_media_entity_form_display_presave() hook doesn't look like it's ever getting called.

modules/lightning_features/lightning_media/tests/src/FunctionalJavascript/MediaReferenceFieldTest.php
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $profile = 'lightning';
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $strictConfigSchema = FALSE;
```
naveenvalecha: Why this is false? Which config schema/metadata is missing? Could we add a todo to remove this after adding the schema/metadata if defining metadata is out of the scope of this issue.

phenaproxima:
EDIT: Moved.

modules/lightning_features/lightning_media/tests/src/FunctionalJavascript/MediaReferenceFieldTest.php
phenaproxima: In this case, it's a flaw in Media Entity Image's plugin settings schema. I can file an upstream patch (or pull one in) but for the short term it's easier to just short-circuit config schema. This test doesn't need the schema to be perfect.

naveenvalecha: I believe A todo with the d.org issue pointing to it would be enough for this PR

phenaproxima: Is there already an issue for this? If so, can you link it here?

naveenvalecha: I'm not sure there's any existing one. Also didn't find any issue after doing a [quick look into the issue queue of media_entity_image](https://www.drupal.org/project/issues/media_entity_image?categories=All) Go ahead and create one

balsama:
This is working great. My only comments are about the UX.

Talked to @phenaproxima and we agreed that adding a view mode "thumbnail" that we use for the selected items _should_ be enough. We might be able to improve on the button verb too (currently "place")

phenaproxima:
Okay! I've added a new 'thumbnail' view mode for all media bundles, and the media reference fields will use it by default.

PR #299, opened 2017-03-16
Generate aggregated Behat configuration dynamically

phenaproxima:
This is auxiliary to #276 -- it allows modules to expose features, contexts, and configuration to Behat by delegating to Drupal the work of aggregating all the config and generating a single, centralized behat.yml, similar to the way Composer Manager works. I did this in Lightning Dev, but the functionality could be split out into a contrib module if it's useful enough. This opens the door to us splitting the tests into their individual subcomponents.

(As a little bonus, I also added a command to our Travis config that runs PHPUnit tests. So we can start using that as desired.)

I'd like @grasmash and @arknoll to review this PR before @balsama approves it, to verify that it won't break BLT (or, if it will, to confirm that the breakage can be fixed easily).


phenaproxima:
It's worth mentioning that this, if merged, will require a few changes to Lightning Project. Nothing major; we simply will need to change the prepare-behat Composer script so that it's aware of the auto-generated file.

phenaproxima:
Looping in @TravisCarden for review.

grasmash:
This does not appear to break BLT integration.

balsama:
Hmm. Looks like there is a legitimate failure on pipelines. Does the pipelines yaml file need updating?

PR #302, opened 2017-03-20
Add unique tags to Behat tests

phenaproxima:
Addresses https://www.drupal.org/node/2862119.

I'd prefer to merge #299 first, and then move the tests into their relevant components, but that needs sign-off from @grasmash (and possibly others) first. In the meantime, this should help improve things too.


balsama:
committed in b24879eb67e450103e2a71afe76ee4de58ea12bf

PR #303, opened 2017-03-20
Provide a facade for manipulating default config

phenaproxima:
This is purely a DX improvement for Lightning, but it opens the door to smoothing out the complex config manipulation that will be involved in splitting the automatic role functionality into its own sub-component.

We used to have functions, like lightning_core_read_config(). Then we moved that stuff into the lightning.config_helper service. This PR makes ConfigHelper extend the core InstallStorage class, which allows it to become a smooth facade around a module's default config, like so:

```php
ConfigHelper::forModule('lightning_core')->getEntity('field_storage_config', 'node.field_meta_tags')
```

This will load the node.field_meta_tags field storage config entity, if it exists -- otherwise it will read it from Lightning Core's default config and return an unsaved entity object, all transparent to the calling code.

It also works for optional config:

```php
ConfigHelper::forModule('lightning_search')->optional()->getEntity('search_api_server', 'database')
```

And it works for simple config (non-entities) as well, except it returns simple config objects rather than entities:

```php
ConfigHelper::forModule('lightning_core')->get('lightning_core.settings')->save()
```

### DIFF
--- a/modules/lightning_features/lightning_core/lightning_core.install
+++ b/modules/lightning_features/lightning_core/lightning_core.install
@@ -6,6 +6,7 @@
  */
 
 use Drupal\Core\Entity\Entity\EntityViewMode;
+use Drupal\lightning_core\ConfigHelper as Config;
 use Drupal\node\Entity\NodeType;
 
 /**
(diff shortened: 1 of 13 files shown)


balsama:
Pipelines failed because of a timeout. Logged to spreadsheet

PR #305, opened 2017-03-20
Split all user role management into a sub-component

phenaproxima:
This addresses https://www.drupal.org/node/2855724.

This here PR introduces a new sub-component of Lightning Core, called Lightning Roles. All the user role functionality that was formerly part of Lightning Core is now in this sub-component. Additionally, other components that provide roles now have those roles as optional config that depends on Lightning Roles. What this adds up to is that if you install Lightning Roles, you will receive all the roles that Lightning has provided until now; if you exclude it, Lightning will do no role management whatsoever. This should make a lot of folks very happy, since role management has proven a bit controversial.

The Lightning Roles sub-component will be installed by default, but you can easily turn it off in a subprofile.


phenaproxima:
For the record, Pipelines failed the latest build because PhantomJS crashed at some point.

balsama:
Awesome. This will make a lot of people happy.

I did run into one issue when testing. On an upgraded site (from 8.x-2.x HEAD to this PR) I get an exception when I try to visit "/admin/config/system/lightning":

    LogicException: Unable to parse the controller name "\Drupal\lightning_core\Controller\SettingsDashboard". in Drupal\Core\Controller\ControllerResolver->createController() (line 120 of core/lib/Drupal/Core/Controller/ControllerResolver.php).

FWIW, I was trying to test to ensure that users who had previously opted out of Lightning roles via the config form would continue to _not_ get roles after the update hook enabled the lightning_roles module. This appeared to work as desired (they did _not_ get roles) but I'm not 100% sure I trust the results because of the exception.

Anyway, we need to figure fix the exception first, then I can confirm. Looks like [Drupal\Core\Controller::createController](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Controller%21ControllerResolver.php/function/ControllerResolver%3A%3AcreateController/8.2.x) is looking for a specific format.

balsama:
Just discussed with @phenaproxima. It would be good if, in the update path, if a user had previously opted out of both of the content roles, the roles subcomponent would not be installed at all.

balsama:
Pipelines failure logged: phantomjs crashed.

balsama:
This works perfectly. If I have neither role selected, the roles module isn't installed at all. If I have only one selected, the module is installed and my preferences are preserved. Perfect.

balsama:
committed in 961bae1a3333f59761a54417618a868adf046f78

PR #307, opened 2017-03-21
Start adding PHPUnit tests

phenaproxima:
Now that we have PHPUnit support, I have written a few tests that take advantage of it. Mostly simple tests of helper classes and other relatively minor bolt-on functionality.

Basically, this PR adds test coverage. Never a bad thing :)

### DIFF
--- a/modules/lightning_features/lightning_core/tests/src/Kernel/Access/AdministrativeRoleCheckTest.php
+++ b/modules/lightning_features/lightning_core/tests/src/Kernel/Access/AdministrativeRoleCheckTest.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\Tests\lightning_core\Kernel\Access;
 
-use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Routing\RouteMatch;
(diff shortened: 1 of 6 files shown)


modules/lightning_features/lightning_core/tests/src/Kernel/Access/AdministrativeRoleCheckTest.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Drupal\Tests\lightning_core\Kernel\Access;
+
+use Drupal\Core\Routing\RouteMatch;
+use Drupal\lightning_core\Access\AdministrativeRoleCheck;
+use Drupal\Tests\token\Kernel\KernelTestBase;
```
naveenvalecha: Could we add a comment above that why we extended this class instead of system KernelTestBase?

phenaproxima: Nice catch. Fixed.

phenaproxima:
Pipelines timed out the last build. @balsama OKed this PR so I'm merging it.

PR #313, opened 2017-03-29
Move experimental components to the experimental branch

phenaproxima:
Just what it says; all experimental stuff should be in the 8.x-2.x-experimental branch, and the "experimental" designation is removed entirely.

Addresses https://www.drupal.org/node/2862124.


balsama:
Committed in 4432246

PR #314, opened 2017-03-29
Make Preview into a first-class component

phenaproxima:
Now that we have an experimental branch, there is no need for the Preview component to have a special experimental designation. This PR removes Lightning's support for experimental components as a special thing, and makes Preview into a normal component that is installed with Lightning as a matter of course.

It's worth noting that tests which are broken by Multiversion are now re-enabled and carry the expect-failure tag so that they are skipped by Travis CI. Thus can the experimental branch continue to have passing builds.


phenaproxima:
Today, Multiversion and Workspace released their first betas! This PR updates both and removes all patches for them, since every patch we were applying had already been committed.

Issue #308, opened 2017-03-22
Patch Drush so that dependencies of base profiles behave the same as normal profiles


danepowell:
Currently Lightning has a hard dependency on Views UI and Database Logging modules. This makes it impossible to uninstall these modules in a production environment running Lightning, which would really be best practice.

Is this something that can be overridden by a subprofile? If not, these dependencies should probably be removed.

phenaproxima:
That sounds OK to me. We could make them dependencies of Lightning Dev so that we will always automatically install them for internal development. @balsama, what do you think?

balsama:
You should be able to uninstall them. Profile dependencies are supposed to be special in that they are bypassed during the check on the module uninstall page.

This previously _was_ a problem with the sub-profile patch. Then it was fixed. Now it appears to be back again. I think we need to fix the problem with the patch. And I think we need to prioritize it - hopefully before 2.0.6 next Wednesday.

balsama:
See comment [256](https://www.drupal.org/node/1356276#comment-11819462) and then [262](https://www.drupal.org/node/1356276#comment-11886317)

danepowell:
Ah your comment about profile dependencies being special makes sense. I was confused too, because I've never run into this problem in the past when uninstalling via the UI. But in this case, I was uninstalling via `core.extensions.yml`, and I got an error during the drupal install / config import process. It seems that the config import process doesn't carve out this same exception.

balsama:
I'm without my computer til tomorrow and thus without phpstorm and a real dev environment. So I can't provide a new patch, but it looks to me like this logic was lost in the patch in comment 269.

```
-      foreach (array_keys($module->required_by) as $dependent) {
-        if ($dependent != $profile && drupal_get_installed_schema_version($dependent) != SCHEMA_UNINSTALLED) {
+      $dependents = array_diff_key($module->required_by, $profiles);
+      foreach (array_keys($dependents) as $dependent) {
+        if (drupal_get_installed_schema_version($dependent) != SCHEMA_UNINSTALLED) {
```

The second `-` line above had been:

`if (!in_array($dependent, array_keys($profiles)) && drupal_get_installed_schema_version($dependent) != SCHEMA_UNINSTALLED) {`

in the previous patch.

balsama:
[...]

PR #316, opened 2017-03-29
Components can't be excluded in subprofile

danepowell:
There was a bug in the existing subprofile patch which I believe broke the excluded_dependencies functionality, meaning you can't currently exclude Lightning components from a subprofile. This updates the subprofile patch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -82,7 +82,7 @@
                 "2765525 - Add AJAX command to add style sheets to CKEditor instances":
                 "https://www.drupal.org/files/issues/2765525-30-8.2.x.patch",
                 "1356276 - Allow profiles to provide a base/parent profile and load them in the correct order":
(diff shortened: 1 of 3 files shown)


balsama:
@danepowell I'm pretty sure this was addressed in #308. Our original assessment that the patch needed to be updated was wrong. See this comment:
https://github.com/acquia/lightning/issues/308#issuecomment-288838410

Can you confirm?

danepowell:
Yeah actually just updating to 2.0.6 (which contains the Drush patch) seemed to fix this. Thanks!

danepowell:
Actually I spoke too soon. Maybe this works for Drush commands, but there are still areas where it fails. However, updating the patch isn't the solution so I'm going to keep this PR closed and open a new issue instead: #317

PR #318, opened 2017-03-31
Verified that Lightning Roles is enabled before attempting Media Image role setup

DevinCarlson:
After updating a profile which extends Lightning, I started seeing errors when installing Lightning Media Image.

In hook_install(), Lightning Media Image attempts to grant the 'media_creator' and 'media_manager' roles the 'access image_browser entity browser pages' permission. However, those roles may not be available as they are optional config which depends on the Lightning Roles module being installed.

It looks like changes were made to the Media, Layout and Workflow modules to accommodate the addition of the Lightning Roles module (which is great btw!), so I made similar changes to Lightning Media Image.

### DIFF
--- a/modules/lightning_features/lightning_media/modules/lightning_media_image/lightning_media_image.install
+++ b/modules/lightning_features/lightning_media/modules/lightning_media_image/lightning_media_image.install
@@ -20,7 +20,9 @@ function lightning_media_image_install() {
 
   // Grants image browser access to the creator content role and the
   // media_creator and media_manager roles.
-  lightning_media_image_update_8004();
+  if (\Drupal::moduleHandler()->moduleExists('lightning_roles')) {
+    lightning_media_image_modules_installed(['lightning_roles']);
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


modules/lightning_features/lightning_media/modules/lightning_media_image/lightning_media_image.install
```diff
@@ -562,18 +564,14 @@ function lightning_media_image_update_8003() {
 }
 
 /**
- * Grants image browser access to media_creator role and creator content role.
+ * Removed in Lightning 8.x-2.07.
```
phenaproxima: There won't be a Lightning 2.07 :) The next release is 8.x-2.10.

modules/lightning_features/lightning_media/modules/lightning_media_image/lightning_media_image.install
```diff
@@ -562,18 +564,14 @@ function lightning_media_image_update_8003() {
...
+ *
+ * Formerly added the 'access image_browser entity browser pages' permission to
+ * the media_creator, media_manager and creator content roles.
```
phenaproxima: It's not clear from Lightning's API, but this is misphrased -- media_creator and media_manager are actual user roles that ship with Lightning Media, whereas "creator" is a meta-role that is managed by the lightning.content_roles service. So this should say "...permission to the media_creator and media_manager roles, as well as the creator content role."

modules/lightning_features/lightning_media/modules/lightning_media_image/lightning_media_image.module
```diff
@@ -12,6 +12,24 @@ use Drupal\lightning_core\OverrideHelper as Override;
 use Drupal\lightning_media_image\Plugin\MediaEntity\Type\Image;
 
 /**
+ * Implements hook_modules_installed().
+ */
+function lightning_media_image_modules_installed(array $modules) {
+  if (in_array('lightning_roles', $modules)) {
+    \Drupal::service('lightning.content_roles')
```
phenaproxima: The grants to media_creator and media_manager should use user_role_grant_permissions(). Only the creator grant should use the lightning.content_roles service.

phenaproxima (changes requested):
Looks pretty good, just need a few changes! :)

danepowell:
I ran into the same bug, can confirm that this PR fixes it. Thanks!

phenaproxima (approved):

balsama:
Timeout on twitter resource on pipelines job.

PR #319, opened 2017-03-31
Update core profile patch to fix config imports

balsama:


danepowell:
Is this patch against 8.2.x? I'm not able to apply it cleanly.

balsama:
No. I didn't back port it. It's against 8.4.x, but also applies to 8.3.x.

The core array syntax patch landed, so it's a pain to backport. And 8.3.0
comes out next Wednesday (along with lightning 2.1.0).

PR #320, opened 2017-04-01
Fix AJAX error when clearing image fields on media items

phenaproxima:
This fixes https://www.drupal.org/node/2865794 and adds a functional regression test.


balsama:
Phantom crashed on pipelines.

balsama:
committed in 9c26fce7108759e663a22d2bcb48cd29cf46c61f

Issue #324, opened 2017-04-03
Lightning + Search API 1.0-beta5


sylus:
Hi there! ^_^

My TravisCI build has suddenly failed and I noticed it was a result of Search API just releasing a new release.

My build is located here: https://travis-ci.org/drupalwxt/wxt/builds/217929831

The errors in particular can be found on this line: https://travis-ci.org/drupalwxt/wxt/builds/217929831#L718

I managed to correct my builds by specifying search_api @ 1.0-beta4 which doesn't have the problem and my build then subsequently passed: https://travis-ci.org/drupalwxt/wxt/builds/217935664

I was curious if anyone else is seeing this?

A subset of the log is given below:

```sh
(search_api) did not specify an instance class. in
/var/www/html/core/lib/Drupal/Component/Plugin/Factory/DefaultFactory.php:79
Stack trace:
#0
/var/www/html/core/lib/Drupal/Core/Plugin/Factory/ContainerFactory.php(17):
Drupal\Component\Plugin\Factory\DefaultFactory::getPluginClass('search_api',
Array, 'Drupal\\views\\Pl...')
#1
/var/www/html/core/lib/Drupal/Component/Plugin/PluginManagerBase.php(84):
Drupal\Core\Plugin\Factory\ContainerFactory->createInstance('search_api',
Array)
#2
/var/www/html/core/modules/views/src/Plugin/views/display/DisplayPluginBase.php(810):
Drupal\Component\Plugin\PluginManagerBase->createInstance('search_api')
#3
/var/www/html/core/modules/views/src/Plugin/views/display/DisplayPluginBase.php(933):
Drupal\views\Plugin\views\display\DisplayPluginBase->getPlugin('cache')
#4
/var/www/html/core/modules/views/src/Plugin/views/display/DisplayPluginBase.php(952):
Drupal\views\Plugin\views\display\DisplayPluginBase->getAllPlugins(true)
#5
/var/www/html/core/lib/Drupal/Core/Plugin/PluginDependencyTrait.php(35):
Drupal\views\Plugin\views\display\DisplayPluginBase->calculateDependencies()
#6 /var/www/html/core/modules/views/src/Entity/View.php(281):
Drupal\Core\Config\Entity\ConfigEntityBase->calculatePluginDependencies(Object(Drupal\views\Plugin\views\display\DefaultDisplay))
#7
/var/www/html/core/lib/Drupal/Core/Config/Entity/ConfigEntityBase.php(346):
[...]

PR #325, opened 2017-04-03
Update Search API to beta5

phenaproxima:
Search API has rolled beta5, which unfortunately has bugs in its update path that broke a bunch of stuff. This PR:

1) Updates to beta5.
2) Applies a patch I wrote to fix the update path problems: https://www.drupal.org/node/2866454
3) Addresses #324, which is caused by invalid default configuration shipped with Lightning Search which is incompatible with Search API beta5. This is also breaking the most recent builds of Lightning Project, which is pulling in beta5.


phenaproxima:
PhantomJS crashed on Pipelines. Logging to the spreadsheet.

PR #326, opened 2017-04-05
Update Panels ecosystem to use layout_discovery instead of layout_plugin

japerry:


sylus:
Yeah this is tripping me up also. I am using lightning v2.1.0 but also bootstrap_layouts. Unfortunately I have to use the newer bootstrap_layouts v5.x to work with the new layout_discovery, however layout_plugin won't disable until panels loses its dependency.

phenaproxima:
@japerry I tried this on my own fork, and it seems good. It just needs, apparently, one change to pass the tests, which is that the test tagged 20e106df in tests/features/layout/panelizer_wizard.feature needs to have a step changed from this:

```
And I place the "Authored by" block into the "middle" panelizer region
```

to this:

```
And I place the "Authored by" block into the "content" panelizer region
```

Once that's done, Travis should pass this (at least when installing from HEAD; not sure about the upgrade path) and we should be able to merge this.

phenaproxima:
Thanks, @japerry.

From my own testing, it looks like we'll also need to update the layout plugin IDs mentioned in modules/lightning_features/lightning_layout/modules/lightning_landing_page/config/install/core.entity_view_display.node.landing_page.full.yml.

modules/lightning_features/lightning_layout/modules/lightning_landing_page/config/install/core.entity_view_display.node.landing_page.full.yml
```diff
@@ -49,7 +49,7 @@ third_party_settings:
         uuid: 5af47416-fa98-4dfb-9761-fc497a47a76f
         label: 'Three Column'
         weight: 0
-        layout: threecol_25_50_25_stacked
+        layout: layout_threecol_25_50_25_stacked
```
naveenvalecha, line 52: There is the change in the layout config name. We require an hook_post_update_name for the sites which are using this layout. Update path requirements needs thoughts from lightning committers cc/  @balsama  @phenaproxima

phenaproxima:
We are going to do this canonically in #344, so I'm closing this PR. Thank you, @japerry, for all you've done to make this possible (which is a great deal more than shown in this PR)!

PR #330, opened 2017-04-07
Randomize user 1 name on install

danepowell:
It's best practice to not use the default "admin" username for user 1, since that's the first thing brute-force attackers will attempt to use. All projects should change this eventually, so let's just automate it and get projects off on the right foot.

### DIFF
--- a/lightning.install
+++ b/lightning.install
@@ -33,6 +33,7 @@ function lightning_install() {
   // Assign user 1 the "administrator" role.
   $user = User::load(1);
   $user->roles[] = 'administrator';
+  $user->setUsername(uniqid('admin-'));
   $user->save();
 
   // Allow authenticated users to use shortcuts.


balsama:
Nice. I wonder if this would annoy anybody though - and I wonder if this oversteps Lightning's province in any way. Also, how would this behave if I were installing via the UI? Like for cloud free users who are just testing? We'd also need to come up with a way to set this to something predictable on our nightly builds.

danepowell:
Yeah it's one of those things that everyone has to do, so I'd really like it to happen automatically on Acquia projects. But it's not big enough to justify its own module and I'm not sure where else to put it 😕

phenaproxima:
I'm not really in favour of this -- it seems like an opinion that Lightning should not have, seeing as how it's used by more people than just Acquia.

My recommendation would be to build a sub-profile on top of Lightning which *does* do this, and use that profile for Acquia projects. It could easily be done in the hook_install() of such a profile.

danepowell:
Sure, and putting this in a subprofile is what we do right now. But I'd prefer if every single project didn't have to build this into their own subprofile. It would be nice if it came out of the box with some Acquia product (even if it's not Lightning).

Apart from the overhead of every project reinventing the wheel, I think we could do more to support best practices. For instance, a vanilla Lightning install on ACE produces an Insight score of something like 80% (mainly because of the default admin username and lack of caching). I think it would be better if Acquia products (Lightning + BLT) installed on an Acquia stack (ACE or ACSF) worked 100% out of the box without requiring additional configuration.

I totally understand if you don't want to make this a Lightning default, but I'm wondering if you have ideas for how else to handle it? What if we put this into the [subprofile template](https://github.com/acquia/lightning/blob/8.x-2.x/templates/profile/install.twig)?

balsama:
After giving this some more thought, I don't think this is something that Lightning should be modifying. The User 1 username only defaults to "admin" if:
* The user _isn't_ installing via the UI
* The user fails to pass the `--account-name` option to `drush site-install`

Actually, defaulting to "admin" is something that only drush does as far as I can tell. So perhaps we should file a patch over there that makes the `--account-name` option more like the `--account-pass` option (where the default is random).

Ultimately though, I think this would be very unexpected behavior for a distribution and our issue queues would be bombarded :)

PR #332, opened 2017-04-10
Implement bulk entity creation API

phenaproxima:
This addresses https://www.drupal.org/node/2672038.

It provides a very basic implementation -- a form, containing a single dropzone, accessible via a "Bulk upload" link on the media list page. You can upload as many files as you like into that dropzone, and when you click "Continue", it will march you through those uploads, one at a time, to set additional required metadata.

There are certainly going to be ways to improve the UX (Inline Entity Form leaps to mind), but this is a good start.

It's also worth mentioning that this rips the NPM/Bower infrastructure out of Lightning. It pulls the dropzone library from the main Packagist; if we need additional NPM libraries in the future, we'll be able to snag them using Asset Packagist. Once we merge this, NPM is no longer a dependency of Lightning, and we'll need to update all documentation in the next release.


sylus:
This looks awesome, thanks so much for this! I was at first quite curious about this "asset packagist" as it seemed a departure from calling npm / bower directly. However seemed to work great for me with my libraries too ^_^

phenaproxima:
Well, we're actually not using Asset Packagist in this PR at all, because the Dropzone library is available on the regular Packagist (for some reason). But yeah, it's great for the future. Our NPM infrastructure is entirely vestigial, and I'm happy to see it go.

phenaproxima:
Due to unresolved problems with packaging the DropzoneJS library, we've decided to alter the scope of this PR a bit. We're now addressing https://www.drupal.org/node/2870740.

What this means is that just about everything already in this PR will be merged...*except* for the bulk upload UI, and any dependency on the DropzoneJS module and JavaScript library. Those will be supplied on drupal.org as a patch to Lightning which provides this functionality as a Lightning Media sub-component, so you can still use it if you want to. Once we've figure out how to properly distribute JS libraries via Composer, we'll be able to merge the patch into Lighting, and that will be that.

balsama:
Committed in 0f57e8f12cd1fd594c8f2cfad79551611a5b673d

Issue #180, opened 2016-10-17
Backlog


balsama:
Infrastructure
- [x] move to selenium
- [x] test new packagist
  - [x] Can we have two packagist? <-- Yes, but only if they don't define the same packages. If they _do_ define the same package, the last packagist to define that package wins, and only its versions are installable.
  - [x] ~~Does it build dependencies~~ <--It does not. Blocked on this
- [x] Don't test so much
- [ ] Improve testing - use lightning project to build test
- [ ] Move away from Phing, use scripts in Lightning Project's composer.json to do build tasks.
- [ ] Start using asset-packagist instead of NPM and Bower
- [ ] Split the tests into suites separated by component

Preview
- [ ] Can't truly resolve conflicts. See [comment on docs here](https://docs.google.com/document/d/10cBoY74LrjVMWSolMwarLNwes4usepHkE-kuE_y7Lk0/edit?disco=AAAAAvYNAvU)
- [x] Authors don't replicate. [Related code in replication](https://github.com/relaxedws/drupal-replication/blob/8.x-1.x/src/Normalizer/EntityReferenceItemNormalizer.php#L38)
- [ ] PHP warning
- [ ] 2819317 Class not found <-- According to the issue, this causes a consistent test failure. Our tests have not been failing, so this is postponed until further notice.
- [ ] URL aliases - prefixes only work for nodes
- [x] URL aliases - overrides don't work
- [ ] Blocks list page lists thing twice
- [x] Core file system issue needs tests
- [ ] Config entities need to be able to be isolated by workspace

Workflow
- [ ] Migration path for content moderation (joe purcell?)
- [ ] Add notifications to state transitions
- [x] Add Diff module and figure out how best to integrate it into Lightning

Media
- [x] Twitter SVGs
- [x] Videos are not constrained within ckeditor (core issue + lightning)
- [ ] Make Entity Embed entities inline-able

Layout
- [ ] Blocks in layouts should be tied to a revision
- [ ] Move IPE to tray without outside in interfering with current system
- [ ] Wizard sucks - document specific steps
- [x] Panelize all core entities (panelizer issues exist already)
- [ ] After switching to new layout, lightning should more clearly show where the regions are
- [x] ["Revert to default" destroys published version of page](https://www.drupal.org/node/2754649)

PR #335, opened 2017-04-11
Make empty layout regions visible

phenaproxima:
This fixes an issue mentioned in our backlog, #180.

Right now, when using Panels IPE, if you are editing a layout that contains any empty regions, the empty regions display a title bar and nothing else. I think this is kinda icky, and it gives very little sense of what the layout looks like, or that the empty regions are buckets into which things may be placed.

This PR adds a bit of CSS to Panels IPE so that empty regions will be displayed as empty grey boxes when the layout is being edited. It looks, like, 1,000 times nicer (IMHO).

### DIFF
--- a/modules/lightning_features/lightning_layout/css/ipe.css
+++ b/modules/lightning_features/lightning_layout/css/ipe.css
@@ -1,0 +1,8 @@
+#panels-ipe-content [data-region-name] [data-droppable-region-name]:empty {
+    min-height: 10em;
+    background-color: #eee;
+    -webkit-box-sizing: border-box;
+    -moz-box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


balsama:
Can we add an additional style to `.ipe-droppable.active.hover`? Right now, it's not clear that I can drop something. There used to be a style built in for that but I think these styles have overridden it.

balsama:
Looking at this a little closer, IPE already provides these styles and a color change on hover, but .ipe-droppable is set to `display: none`. `.ipe-droppable.active` is set to `display: block`.

So can we just set `.ipe-droppable` to `display: block` and be done with it?

PR #336, opened 2017-04-11
Take advantage of Entity Browser widget contexts for upload validation

phenaproxima:
Remember when we added upload validation to our FileUpload entity browser widget? That was a pretty great win. Turns out, however, that Entity Browser's maintainers (specifically @mortenson)  have been working to bring such things into Entity Browser. In the name of shedding code and maintenance duties from Lightning, I am fully cheerleading the improvements in EB and pushing for us to adopt the things they come up with.

To that end, Entity Browser now has a patch that can pass contextual information into their Upload widget to do upload validation. Our own widget is quite similar, and we can take advantage of that patch (RTBCing it in the process, since we test it quite thoroughly) to do our file validation rather than having to rig everything up ourselves.

This PR integrates that patch into Lightning and removes the code in our FileUpload widget which was previously doing those duties.


phenaproxima:
The Entity Browser issue in question is here: https://www.drupal.org/node/2761833

phenaproxima:
Noting here that @mortenson deserves some commit credit on this one, since he wrote https://www.drupal.org/node/2867473, which does essentially the same thing as this PR.

balsama:
I had some trouble when manually testing. It doesn't seem related to the update path as I was just able to reproduce on a fresh install.

1. Add a media reference field to basic page
1. Select at least two Bundles under Reference type (e.g. Image and Instagram - if you select a single bundle, the filter works fine)
1. Create at least one media image entity
1. Create a new basic page

**Expected result:**  
The media image entity is available in the media reference entity browser

**Actual result:**  
The media reference entity browser reads _"There are no media items to display"_

If I had to take a stab, I'd guess the view if trying to find content that is _both_ Image _and_ Instagram, rather than _or_.

balsama:
Aha. It looks like the "Allow multiple values" checkbox just needs to be checked under the "More" section of the Media: Bundle contextual filter. Here's a quick export of the view after I made that change if it helps:
https://bin.acquia.com/?6cc6fdb988150253#6WzTBz6kALaEAXvHw2JVVvms0wMyHB3MdY3ahUPHjLQ=

phenaproxima:
@balsama, I have made the appropriate changes. Can you test again?

balsama:
Committed in 556c7ec4e34604e8349b67254f54e35611a487cb

PR #340, opened 2017-04-17
Loosen the implicit dependency on Field UI

phenaproxima:
This addresses #327 by wrapping some of the logic in lightning_core_entity_type_alter() in a check to ensure that the Field UI module exists.

### DIFF
--- a/modules/lightning_features/lightning_core/lightning_core.module
+++ b/modules/lightning_features/lightning_core/lightning_core.module
@@ -13,9 +13,6 @@ use Drupal\lightning_core\Element as ElementHelper;
 use Drupal\lightning_core\Entity\EntityFormMode;
 use Drupal\lightning_core\Entity\EntityViewMode;
 use Drupal\lightning_core\Entity\Role;
-use Drupal\lightning_core\Form\EntityDisplayModeAddForm;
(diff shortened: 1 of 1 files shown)


modules/lightning_features/lightning_core/lightning_core.module
```diff
@@ -30,11 +27,13 @@ function lightning_core_entity_type_alter(array &$entity_types) {
 
   Override::entityForm($entity_types['user_role'], RoleForm::class);
 
-  Override::entityForm($entity_types['entity_view_mode'], EntityDisplayModeAddForm::class, 'add');
-  Override::entityForm($entity_types['entity_view_mode'], EntityDisplayModeEditForm::class, 'edit');
+  if (\Drupal::moduleHandler()->moduleExists('field_ui')) {
```
naveenvalecha, line 30: Nice. Could we also remove the hard dependency of field_ui from lightning_core.info.yml file in this patch?

balsama:
@naveenvalecha Do you mean from lightning.info.yml? (It's not a dependency in lightning_core.info.yml)

The dependency in the lightning.info.yml is a profile dependency, so it can be uninstalled or excluded via a sub-profile.

PR #342, opened 2017-04-18
ModerationAwareEntityFormTrait is not very smart

phenaproxima:
This PR addresses https://www.drupal.org/node/2867465.

In a nutshell, the problem is that our NodeForm::actions() method doesn't bother to check if the entity being worked on is, in fact, moderatable. So it goes and mucks with the buttons even if it shouldn't. This adds a check to prevent that, plus a regression test.


balsama:
Committed in bcabb34cc040dfcb0663da729802bfb22923762c

PR #337, opened 2017-04-12
feat(layout): Switch to layout discovery and deprecate layout_plugin

sylus:
This is based off of https://github.com/acquia/lightning/pull/326 but I wasn't able to add to push to that P.R. so opened this one. Wanted the patch so can leverage it in my composer.json. ^_^ Feel free to close though :)

I corrected the changed entries to lightning_landing_page and also fixed up some of the drush make issues since the old patch was calling 7.x-4.x branch. I confirmed the drush make works properly.

I did also have to update ctools as it looks like the dev version panelizer requires `Drupal\ctools\Context\AutomaticContext` and I was having the following issue with search api calling rendered item(s) which in turns calls panelizer needing this context.

```sh

Error: Class 'Drupal\ctools\Context\AutomaticContext' not found in /var/www/html/modules/contrib/panelizer/src/PanelizerEntityViewBuilder.php on line 214 #0 /var/www/html/modules/contrib/panelizer/src/PanelizerEntityViewBuilder.php(381): Drupal\panelizer\PanelizerEntityViewBuilder->getEntityContext(Object(Drupal\node\Entity\Node))
#1 /var/www/html/modules/contrib/panelizer/src/PanelizerEntityViewBuilder.php(356): Drupal\panelizer\PanelizerEntityViewBuilder->buildPanelized(Object(Drupal\node\Entity\Node), Object(Drupal\panels\Plugin\DisplayVariant\PanelsDisplayVariant), 'default', 'en')
#2 /var/www/html/modules/contrib/panelizer/src/PanelizerEntityViewBuilder.php(274): Drupal\panelizer\PanelizerEntityViewBuilder->buildMultiplePanelized(Array, Array, 'default', 'en')
#3 /var/www/html/modules/contrib/search_api/src/Plugin/search_api/datasource/ContentEntity.php(807): Drupal\panelizer\PanelizerEntityViewBuilder->view(Object(Drupal\node\Entity\Node), 'default', 'en')
#4 /var/www/html/modules/contrib/search_api/src/Plugin/search_api/processor/RenderedItem.php(271): Drupal\search_api\Plugin\search_api\datasource\ContentEntity->viewItem(Object(Drupal\Core\Entity\Plugin\DataType\EntityAdapter), 'default')
#5 /var/www/html/modules/contrib/search_api/src/Item/Item.php(248): Drupal\search_api\Plugin\search_api\processor\RenderedItem->addFieldValues(Object(Drupal\search_api\Item\Item))
[...]

PR #344, opened 2017-04-18
Switch to Panels/Panelizer 4.x

phenaproxima:
Well, we finally did it. Wow. All tests pass, and there is no need for any patches!

This replaces #337 and #326, and addresses https://www.drupal.org/node/2870521.

In a nutshell, this PR updates Lightning to the 8.x-4.x branch of the Panels ecosystem. Layout Plugin is gone forever, replaced by the experimental (but more stable than Layout Plugin) core Layout Discovery module.


phenaproxima:
Let's give @japerry and @sylus commit credit on this one, since they both filed PRs and worked to address the same issue.

sylus:
This is amazing big thanks to @phenaproxima + @japerry and to all the panels / ctools maintainers! Pretty big deal! ^_^

phenaproxima:
You're welcome! Happy to do it. Onward and upward!

balsama:
@phenaproxima I haven't attempted to debug this at all, but I did get an EntityStorageException when I tried to test this manually during database updates (it specifically appears to be the panelizer update). In my manual test, I had a BUNCH of landing pages with various layouts and various overrides. But I'm not sure if that is a factor or not yet.

Output of drush update-db:

```
$ ../bin/drush updb
The following updates are pending:

lightning_layout module : 
  8009 -   Replaces Layout Plugin with Layout Discovery. 

panels module : 
  8401 -   Uninstalls Layout plugin, then enables Layout Discovery. 

page_manager module : 
  Rename layout machine names in config entities to match layout discovery's default layouts.

panelizer module : 
  Rename layout machine names in config entities to match layout discovery's  default layouts.

Do you wish to run all pending updates? (y/n): y
Performing lightning_layout_update_8009                                                                                    [ok]
Performing panels_update_8401                                                                                              [ok]
Post updating page_manager                                                                                                 [ok]
Post updating panelizer                                                                                                    [ok]
Failed: Drupal\Core\Entity\EntityStorageException: The theme implementations may not be rendered until all modules are     [error]
loaded. in Drupal\Core\Entity\Sql\SqlContentEntityStorage-&gt;save() (line 770 of
/Users/adam.balsam/Sites/lightning/build/docroot/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).
Cache rebuild complete.                                                                                                    [ok]
Finished performing updates.
```

phenaproxima:
There's always something...

@balsama, I suspect that you having landing pages is *definitely* a factor. Why the hell it would trigger the theme manager (which is what throws the exception) is anybody's guess at the moment. @japerry, can you comment?

balsama:
The minimum steps to reproduce the above seem to be:

1. Create a landing page
1. Place at least one item in the layout via the IPE (I placed the UUID content block)
1. Save the layout as _Custom_
1. Update codebase
1. Run update-db

Result: exception noted above during panelizer update:

```
Failed: Drupal\Core\Entity\EntityStorageException: The theme implementations may not be rendered until all modules are     [error]
loaded. in Drupal\Core\Entity\Sql\SqlContentEntityStorage-&gt;save() (line 770 of
/Users/adam.balsam/Sites/lightning/build/docroot/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage
```

phenaproxima:
I have a suspicion that this is a conflict with Search API. Can you quickly test it with Search API (and our component for it) turned off?

balsama:
Possibly a second problem. Seems likely to be related to the first, but I wanted to document it.

Regions of existing landing pages are "rearranged" after update - new content using the same layout appear fine.

1. Create a landing page and select Two Column for the layout
1. Place a block in the left and and right regions (I placed content UUID in the left and content Title in the right
1. Save the layout as custom
1. Update codebase and run database updates (note that you will produce the exception noted above, but it does not render the site inoperable)
1. Visit the landing page
1. Note that while the blocks are still assigned to the correct regions (left and right) the left and right regions appear stacked on top of each other now (on a large screen).
1. Create a new landing page with the same layout and place the same blocks in the same regions
1. Note that the left and right regions are next to each other as expected in the new landing page. 

Will test with lightning_search and search_api uninstalled now.

balsama:
@phenaproxima Indeed. Running the same steps in https://github.com/acquia/lightning/pull/344#issuecomment-295040838 with search_api_db, search_api, and lightning_search uninstalled did NOT reproduce the exception.

phenaproxima:
Interestingly, I have been unable to reproduce it locally even with Search API enabled. I followed the steps to reproduce, using Lightning 2.1.0. I had a page with the simple one-column layout and the UUID content block. It failed to cause an exception during the update. Which version of Lightning, and which layout, are you using?

balsama:
I'm using HEAD of 8.x-2.x - not 2.1.0. Could that be related?

(I guess I'll test)

php 7.1 in my dev env too which is another difference but can't imaging that's involved.

phenaproxima:
Doesn't seem very likely, but I'll try with that.

balsama:
And I just tried with tags/2.1.0 as a starting point and was able to reproduce again.

It's late, so let me include my entire bash history in the STR from the last steps too - just so we're not missing something (read: to confirm I'm not doing a stupid), prefaced with a `$`:

1. $ git checkout tags/2.1.0
1. $ sudo phing destroy && composer install && phing install
1. Create a landing page
1. Place at least one item in the layout via the IPE (I placed the UUID content block)
1. Save the layout as custom
1. $ git checkout phenaproxima-update-panels
    * Branch phenaproxima-update-panels was previously created via:
        1. git checkout -b phenaproxima-update-panels 8.x-2.x
        1. git pull https://github.com/phenaproxima/lightning.git update-panels
1. $ composer install
1. $ cd docroot/
1. $ ../bin/drush updb

Result: EntityStorageException: The theme implementations may not be rendered until all modules are loaded

phenaproxima:
It's late, I'm going to sleep.

Here's where I'm at. I am pretty sure the two problems are unrelated:

1) I'm sure this is a conflict with Search API. Trying to actually work around Search API is hard, but there might be a hack. Try adding the following line to the end of lightning_layout_update_8009():

```
\Drupal::moduleHandler()->loadAll();
```

I can't guarantee that'll fix it, but it might have some sort of effect. In the meantime, I seem to be unable to reproduce the problem. We'll try tomorrow.

2) I think this is due to bad file propagation after the code base update. Take a look at docroot/profiles/lightning/modules/contrib/panels/layouts. You should see only these directories: http://cgit.drupalcode.org/panels/tree/layouts?h=8.x-4.x. But I suspect you'll see a whole lot more -- which won't have the layout_ prefix. Not only that, but they'll have distinctly different templates which don't load in CSS correctly, thus causing them to appear "rearranged". I'm almost certain that this is the result of Panelizer's post-update hook failing to map the old layout ID to its new one.

3) It would help if I could see the full output of drush updb. Can you post that (tomorrow is fine)?

balsama:
Hopefully this last response doesn't stir you 😴  Just wanted to get it down now. Talk tomorrow. Thanks.

1. This didn't seem to have any effect. Same exception during updb
1. Yes! That directory has eight additional directories in it - none of which start with `layout_`. Good call. So, it's not just a matter of our use of rsync not cleaning those dirs up? The layouts don't know to use the new, erm, layout?
1. Full output of drush updb:

```
$ ../bin/drush updb
The following updates are pending:

lightning_layout module : 
  8009 -   Replaces Layout Plugin with Layout Discovery. 

panels module : 
  8401 -   Uninstalls Layout plugin, then enables Layout Discovery. 

page_manager module : 
  Rename layout machine names in config entities to match layout discovery's default layouts.

panelizer module : 
  Rename layout machine names in config entities to match layout discovery's  default layouts.

Do you wish to run all pending updates? (y/n): y
Performing lightning_layout_update_8009                                                                                    [ok]
Performing panels_update_8401                                                                                              [ok]
Post updating page_manager                                                                                                 [ok]
Post updating panelizer                                                                                                    [ok]
Failed: Drupal\Core\Entity\EntityStorageException: The theme implementations may not be rendered until all modules are     [error]
loaded. in Drupal\Core\Entity\Sql\SqlContentEntityStorage-&gt;save() (line 770 of
/Users/adam.balsam/Sites/lightning/build/docroot/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).
Cache rebuild complete.                                                                                                    [ok]
Finished performing updates.
```

kekkis:
The exception is thrown originally from `Drupal\Core\Theme\ThemeManager::render()`:
`core/lib/Drupal/Core/Theme/ThemeManager.php:139` says

    if (!$this->moduleHandler->isLoaded() && !defined('MAINTENANCE_MODE')) {
      throw new \Exception('The theme implementations may not be rendered until all modules are loaded.');
    }

So one would possibly think that setting the site into maintenance mode might be required as a work-around for the `updb` to succeed.

phenaproxima:
I think that defining MAINTENANCE_MODE will hack around the problem with possibly unexpected results. I decided instead to add a post-update hook to Lightning Layout which force-reloads all modules (using the reload() method of the module handler). In my local testing, this bypasses the problem. Very tricky!

balsama:
Perfect. Confirmed that those two commits fixed both issues. Holding off commit to see if there is any movement on the panels side before core releases.

balsama:
Fixed in c7fd4d3060a5ffdcd0fe124cfd61a4c2b19d6d6b

PR #345, opened 2017-04-18
Issue #343: Fixed Lightning Media modifying configuration during sync.

### LINKED ISSUE #343: Cannot manage configuration provided by Lightning Media
Currently, you cannot manage the configuration provided by Lightning Media, and in fact you can't really use the module at all if you want to use the core configuration system.

The cause is described in this core issue: https://www.drupal.org/node/2870864

In a nutshell: a standard configuration management workflow would be to export your entire site's configuration (including default config provided by Lightning Media at install time) to a directory in your VCS, such as /config/default. The problem is that you can't subsequently import this configuration without error, as described in the core issue.

This is pretty clearly a core bug, not a bug with Lightning, but I want to raise it here since it will primarily impact Lightning users and potentially impact adoption.

danepowell:
Fixes #343

### DIFF
--- a/modules/lightning_features/lightning_core/lightning_core.install
+++ b/modules/lightning_features/lightning_core/lightning_core.install
@@ -11,6 +11,12 @@ use Drupal\Core\Entity\Entity\EntityViewMode;
  * Implements hook_install().
  */
 function lightning_core_install() {
+  // Don't modify configuration if it is currently being synced.
+  // @see https://www.drupal.org/node/2870864#comment-12044705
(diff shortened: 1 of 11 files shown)


phenaproxima:
I have RTBCed the corresponding d.o issue. https://www.drupal.org/node/2870864

phenaproxima:
So this is addressed more thoroughly by #360 -- we'll merge that one. Thanks, @danepowell, for your work here, and for uncovering the issue!

PR #352, opened 2017-04-25
Show the bundle filter if the contextual filter value is an exception

### LINKED ISSUE #351: YieldToArgumentTrait prevents the Bundle filter from ever being shown.
Reviewing the logic at [YieldToArgumentTrait.php](https://github.com/acquia/lightning/blob/8.x-2.x/modules/lightning_features/lightning_core/src/YieldToArgumentTrait.php), it looks like there is no case where the "bundle" exposed filter will ever be shown for the Media browser.

If a contextual filter is configured for the view but a fallback value is used, the exposed filter should be shown. This is not the current behavior, which you can see by visiting /admin/structure/views/view/media/edit/entity_browser_1 and seeing that there is no exposed filter present in the View preview, as the contextual filter has a default value of "all".

mortenson:
In \Drupal\lightning_core\YieldToArgumentTrait, the "bundle" exposed filter is hidden if the contextual filter it yields to has a value. As the Media View is configured to have a fallback value ("all") for its contextual filter, it always appears as if that argument has a value. We should add an extra check in YieldToArgumentTrait that shows the exposed filter if the argument it yields to is returning a value that equals its "exception" value (i.e. the value that shows all results if present, "all").

This resolves #351.

### DIFF
--- a/modules/lightning_features/lightning_core/src/YieldToArgumentTrait.php
+++ b/modules/lightning_features/lightning_core/src/YieldToArgumentTrait.php
@@ -28,7 +28,7 @@ trait YieldToArgumentTrait {
     $value = $argument->getValue();
 
     $key = $this->options['expose']['identifier'];
-    $form[$key]['#access'] = is_null($value);
+    $form[$key]['#access'] = is_null($value) || $argument->isException($value);
   }
 
   /**


phenaproxima:
We'll need a Behat test to prove this is fixed, but it looks great to me. Thanks, @mortenson!

balsama:
Thanks Sam! Merged with a test in e1a733bafb3afc46726452b234fe53fac095ba25

Issue #357, opened 2017-04-30
Lightning Media browser filefield_paths error


luisvivasb:
I'm tried to use the filefield_paths module to transliterate the file filenames but it doesn't work from Lightning Media browser.

If I add a new Image media from the media admin page its works perfect (media/add/image), the image file name is transliterate and save the file into the custom directory config for that.

To reproduce this:
Install and activate the modules:
File (Field) Paths (8.x-1.x-dev): https://www.drupal.org/project/filefield_paths
Pathauto (8.x-1.0): https://www.drupal.org/project/pathauto

Go to Estructure, Media Bundles and edit the Image field options (admin/structure/media/manage/image/fields):
Active the "Enable File (Field)" I custom the File path rute to `media/[date:custom:Y]-[date:custom:m]/img` in the File name left the default value `[file:ffp-name-only-original].[file:ffp-extension-original]`
On File name options active the "Cleanup using Pathauto" and "Transliteration".
Save.

Goto Content, Media (admin/content/media) and add new content type Image (media/add/image), please select a Image from your computer with symbols like `(,&$.ñ¡!? )` and save. This step work fine you can see the image and verify the folder when where saved (admin/content/files).

Now goto Add new content like Article or Basic Page (node/add), use the Text Editor Media Browser to add image to body content, please upload a new Image from your computer with symbols like `(,&$.ñ¡!? )` and save it.

Now you can see from Admin Media (admin/content/media) that the new image don't show.

Over the Files Admin (admin/content/files) you can compare the images, and see that the image uploaded from the Media Browser was saved with the filename without trasliteration and was not saved in the custom directory.

But if you check your public directory you see the file in the correct custom folder and with the transliteration name.

On my test my image filename was: `Lá Mórüta y su eñe Bol&so)(*&^ new dos.JPG`
The correct folder and filename: `public://media/2017-04/img/la-moruta-y-su-ene-bolso-new-dos.JPG`
The wrong result: `public://Lá Mórüta y su eñe Bol&so)(*&^ new dos.JPG`

Thanks for the help!

balsama:
Migrated to https://www.drupal.org/project/lightning/issues/2924489

PR #358, opened 2017-05-01
Update FileUpload.php

luisvivasb:
The $file->setPermanent() is already call when you use the file in media entity. If you call this function at this point you can't use tokens or filefield_paths to manipulate the image name and folder to be saved.

#357 You can see the problem with filefield_paths

### DIFF
--- a/modules/lightning_features/lightning_media/src/Plugin/EntityBrowser/Widget/FileUpload.php
+++ b/modules/lightning_features/lightning_media/src/Plugin/EntityBrowser/Widget/FileUpload.php
@@ -104,8 +104,8 @@ class FileUpload extends EntityFormProxy {
       $entity,
       MediaHelper::getSourceField($entity)->entity
     );
-    $file->setPermanent();
-    $file->save();
+    //$file->setPermanent();
+    //$file->save();
     $entity->save();
 
     $selection = [


modules/lightning_features/lightning_media/src/Plugin/EntityBrowser/Widget/FileUpload.php
```diff
@@ -104,8 +104,8 @@ public function submit(array &$element, array &$form, FormStateInterface $form_s
       $entity,
       MediaHelper::getSourceField($entity)->entity
     );
-    $file->setPermanent();
-    $file->save();
+    //$file->setPermanent();
```
phenaproxima, line 107: I can't merge this with unexplained, commented-out code in there. If this makes it work (and apparently it still passes the tests, so hey, alright!), then we need a comment here explaining why the file is not saved.

phenaproxima (changes requested):
Looks useful, but needs a bit of refinement...

phenaproxima:
This looks like a nice win. I think we're going to need a test of it, though. That's going to be fairly involved, since we'll likely have to write a few new Behat step definitions. I can probably open a feature branch and work on this there, if you'd prefer. Because I'd definitely like to get this committed.

PR #361, opened 2017-05-08
Verify Patched constraints

balsama:
We've made the mistake a few times of adding a patch to a module without pinning it to a specific release. Then, when the patch gets committed and the maintainer makes a new release, builds fail because the patch we define no longer applies.

This PR adds a composer script that validates all patched dependencies have pinned constraints and runs that script in the .travis.yml script.


balsama:
Closing until we work out travis's problem.

balsama:
Fixed the command so that it runs within docroot

src/Composer/PatchedConstraint.php
```diff
@@ -0,0 +1,85 @@
...
+   *
+   * @param \Composer\Script\Event $event
+   *   The script event.
+   *
+   * @return bool
+   *   False if there are unpinned patched dependencies.
+   */
+  public static function execute(Event $event) {
+    $root_package = $event->getComposer()->getPackage();
+    $patched_dependencies = self::getPatchedDependencyConstraints($root_package);
```
phenaproxima: Nitpick: We should generally use static, not self.

src/Composer/PatchedConstraint.php
```diff
@@ -0,0 +1,85 @@
...
+    $error = [];
+
+    /** @var \Composer\Package\Link $package */
+    foreach ($patched_dependencies as $package) {
+      if (self::packageIsUnpinned($package)) {
```
phenaproxima: We should favor static over self.

src/Composer/PatchedConstraint.php
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * @param \Composer\Package\Link $package
+   *   The package to check.
+   *
+   * @return bool
+   *   True if the constraint appears to be unpinned.
+   */
+  protected static function packageIsUnpinned(\Composer\Package\Link $package) {
```
phenaproxima: This method is missing a description in the doc comment. Also, can the type hint be shorted to Link and used at the top of the file?

src/Composer/PatchedConstraint.php
```diff
@@ -0,0 +1,85 @@
...
+    if ($package->getTarget() == 'drupal/core') {
+      // Bail out if the patched package is drupal/core since we release with
+      // each version of core and always ensure core patches still apply.
+      return false;
+    }
+    $constraint = $package->getPrettyConstraint();
+    if ((!is_numeric(substr($constraint, 0, 1))) || (strpos($constraint, '|') !== FALSE) || (strpos($constraint, '*') !== FALSE)) {
```
phenaproxima: Wow, this is complex. Can we use a regex instead? It would be relatively simple -- all we'd need to do is this: /^[0-9\\.]+(-alpha|beta|rc[0-9]+)?$/

composer.json
```diff
@@ -49,7 +49,8 @@
         ],
         "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
         "package": "Acquia\\Lightning\\Composer\\Package::execute",
-        "release-version": "Acquia\\Lightning\\Composer\\ReleaseVersion::execute"
+        "release-version": "Acquia\\Lightning\\Composer\\ReleaseVersion::execute",
+        "verify-patched-constraints": "Acquia\\Lightning\\Composer\\PatchedConstraint::execute"
```
phenaproxima, line 55: I think this should be a pre-commit hook, but we can bikeshed on that later.

phenaproxima (changes requested):
This is really nice. I mostly have nits.

balsama:
Agree with everything including moving this to a pre-commit hook:
* Moved to git-hooks/pre-commit which will now fail if verify-patched-dependencies returns a non-zero
* Add an additional composer script (symlink) that automatically symlinks .git/hooks/pre-commit to git-hooks/pre-commit 
* Updated composer.json so that composer symlink is run on install and update

phenaproxima (changes requested):
I don't think we need the Symlink thing to be a full-fledged Composer script. We should be able to just do this as the symlink script: ln -s -f git-hooks/pre-commit .git/hooks/pre-commit

balsama:
Fair enough. I'd prefer to add it as a phing target so that we're not duplicating the same code in the install and update sections.

phenaproxima (approved):

phenaproxima:
This looks great. I think we should merge it.

PR #362, opened 2017-05-08
"Create new revision" checkbox hidden for nodes that are not using workbench moderation

balsama:
See D.O https://www.drupal.org/node/2876698. Added test.

### DIFF
--- a/tests/features/workflow/moderation_states.feature
+++ b/tests/features/workflow/moderation_states.feature
@@ -116,3 +116,4 @@ Feature: Workflow moderation states
     When I visit "/deft-zebra"
     And I click "Edit"
     Then I should see "Create new revision"
+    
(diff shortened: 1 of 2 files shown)


phenaproxima:
Looks good to me, and Travis seems happy with it. Let's merge it!

tests/features/workflow/moderation_states.feature
```diff
@@ -103,3 +103,17 @@ Feature: Workflow moderation states
...
+    Given node_type entities:
+      | type          | name          |
+      | not_moderated | Not Moderated |
+    And not_moderated content:
+      | title       | path       |
+      | Deft Zebra  | /deft-zebra |
+    And I am logged in as a user with the administrator role
+    When I visit "/deft-zebra"
+    And I click "Edit"
+    Then I should see "Create new revision"
```
phenaproxima, line 118: I have only one question -- is there a step like "Then I should see a 'Create new revision' checkbox", so that this is a little less ambiguous? If not, no big deal, I still think the test proves this works, but it's worth being more specific if we can.

balsama:
Fair point. I could assert that the checkbox is checked or unchecked, but that's not really what we want. No better step definition exists, so this will do for now. Thanks!

balsama:
Commited in 0b2b20650c1113d2a6aaf7cd21be9ae51e1ef763

PR #363, opened 2017-05-12
Entity browsers for media reference should tell the user how many items they can choose

phenaproxima:
This PR addresses UX feedback that is fixed by this patch I filed against Entity Browser: https://www.drupal.org/node/2877751


balsama:
This is nice!

One comment. This PR includes an update to Search API (that includes a DB update) that's out of scope. Can we remove it? Looks like it also has an update to Media Entity Instagram.

PR #365, opened 2017-05-12
Fixed stale configuration exports.

danepowell:
At some point, the Entity Browser and Media modules were updated, and these updates brought with them some config schema changes. However, these schema changes were never captured in Lightning's exported configuration.

For instance, Entity Browser added an [auto_select setting](https://www.drupal.org/node/2822009) in 1.0-beta4, but when Lightning updated Entity Browser to 1.0-beta4 it didn't re-export the Entity Browser config to capture this change.

In the future, it's important when updating modules to follow a process such as the one [suggested by BLT](http://blt.readthedocs.io/en/stable/readme/configuration-management/#updating-core-and-contributed-modules) to capture configuration changes. I think these are relatively harmless but it could cause more serious and hard to diagnose problems.

### DIFF
--- a/modules/lightning_features/lightning_media/config/install/entity_browser.browser.media_browser.yml
+++ b/modules/lightning_features/lightning_media/config/install/entity_browser.browser.media_browser.yml
@@ -1,6 +1,8 @@
 langcode: en
 status: true
 dependencies:
+  config:
+    - views.view.media
   module:
     - lightning_media
     - views
@@ -8,11 +10,11 @@ name: media_browser
 label: 'Media browser'
 display: iframe
 display_configuration:
-  path: /media-browser
   width: 100%
   height: '640'
   link_text: 'Select entities'
   auto_open: true
+  path: /media-browser
 selection_display: no_display
 selection_display_configuration: {  }
 widget_selector: tabs
@@ -20,9 +22,10 @@ widget_selector_configuration: {  }
 widgets:
   134808d9-d854-4a0b-8699-d5eba006b8b7:
     settings:
-      submit_text: Place
       view: media
       view_display: entity_browser_1
+      submit_text: Place
+      auto_select: false
     uuid: 134808d9-d854-4a0b-8699-d5eba006b8b7
     weight: -10
     label: Library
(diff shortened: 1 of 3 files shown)


phenaproxima:
This is a really nice catch, and definitely a good-sized hole in our release process. I think what I'd like to do is merge this PR, then write a script that runs as part of our release-making infrastructure which will re-export and overwrite Lightning's default configuration. That's a bigger project that I'll take up today, though, and in the meantime, this PR is "RTBC" from me :)

phenaproxima:
So I wrote a Composer script which can update Lightning's default configuration with the latest from Drupal, then ran smack into Trump's wall when I realized that this approach _definitely_ will not work for Lightning. Why? Because although a lot of our config is used as shipped, a lot of it is also heavily modified in complicated ways at install time, during update hooks, or just in the course of general site building, depending on what components of Lightning are present and how they're configured. It's impossible for us to re-export the config, as Lightning installs it, and call it a day. The problem(s) manifest in the form of many sudden hard dependencies, which really need to be soft dependencies, showing up in the exported config.

I'm not really sure what approach to take from here. Maybe we could whitelist a bunch of config that Lightning does not touch, and only automatically re-export that? It might work, but it would be clumsy and not very future-proof.

I'm open to ideas...

balsama:
Totally agree with @phenaproxima 's conclusion. Configuration exported by the system isn't necessarily the same config that every install of Lightning would want because we allow components to be excluded.

But if Lightning were to export it's completed config on each commit into an inactive/unused by Drupal directory, a difference between commits would likely mean that there is something that should be changed in Lightning's provided config.

So how about we automate a config export by Lightning on each update of dependencies. It would look something like this:

* composer update would trigger:
  1. A fresh site install
  1. A config export into a fixture directory

Of course, this would only affect those building Lightning from source. Any difference in the unused exported config would trigger a manual followup of the related config used by Lightning.

balsama:
Meanwhile, this PR is still good to go, correct? We can move this conversation elsewhere if so.

danepowell:
I agree, the way to track this is probably via diffs, rather than just relying on a static dump of configuration. Basically:

1. Start with a known "clean" install of Lightning
2. Temporarily export the site configuration (`drush cex`)
3. Use git to track the exported config, or copy it to a new directory for safekeeping
3. Run composer update
4. Export the site configuration a second time, and compare the newly-exported configuration to the original, either via `git diff` or `diff`.
5. If there are any changes, manually correlate them to configuration that's provided by Lightning modules and update the corresponding config files.

It's a bit of a pain, but I'm not sure how else you'd accomplish this.

balsama:
I opened up #367 and #366 as followups.

Issue #367, opened 2017-05-16
Implement method to identify config that has become outdated


balsama:
Followup of #365. From that issue:

@phenaproxima:
> So I wrote a Composer script which can update Lightning's default configuration with the latest from Drupal, then ran smack into Trump's wall when I realized that this approach definitely will not work for Lightning. Why? Because although a lot of our config is used as shipped, a lot of it is also heavily modified in complicated ways at install time, during update hooks, or just in the course of general site building, depending on what components of Lightning are present and how they're configured. It's impossible for us to re-export the config, as Lightning installs it, and call it a day. The problem(s) manifest in the form of many sudden hard dependencies, which really need to be soft dependencies, showing up in the exported config.

@balsama:
> If Lightning were to export it's completed config on each commit into an inactive/unused by Drupal directory, a difference between commits would likely mean that there is something that should be changed in Lightning's provided config.
> 
> So how about we automate a config export by Lightning on each update of dependencies. It would look something like this:
> 
> composer update would trigger:
> * A fresh site install
> * A config export into a fixture directory
> 
> Any difference in the unused exported config would trigger a manual followup of the related config used by Lightning.

@danepowell 

> I agree, the way to track this is probably via diffs, rather than just relying on a static dump of configuration. Basically:
> 
> 1. Start with a known "clean" install of Lightning
> 1. Temporarily export the site configuration (drush cex)
> 1. Use git to track the exported config, or copy it to a new directory for safekeeping
> 1. Run composer update
> 1. Export the site configuration a second time, and compare the newly-exported configuration to the original, either via git diff or diff.
> 1. If there are any changes, manually correlate them to configuration that's provided by Lightning modules and update the corresponding config files.

danepowell:
I forgot a step: after composer update, run drush updb

balsama:
@danepowell Thanks. I think we're going to take a slightly different approach, and compare config exported from a fresh install of Lightning. So each time we install using our development target, we'll regenerate the config. The config will be stored in our repo (somewhere within the tests directory so it's clear that it's not to be used) and we'll use git to compare differences. We'll strip UUIDs out of the stored config so as to not create commit cruft.

90% of the time, there will be no difference. But cex is cheap! 👍 When there is a difference (e.g. after we've run `composer update`), we'll manually correlate and resolve that difference in Lightning's config.

We'll see how it works. Working on a PoC now.

sylus:
[...]

PR #368, opened 2017-05-16
Continuously generate config snapshots for sanity checking purposes

phenaproxima:
This PR implements @balsama's idea in #367. Every time we run our standard 'install' Phing target, all active config will be exported into the local tests directory so that we can compare it against our commit history and see what, if anything, has changed. This will give us a clearer picture into how we might need to adjust our default config to keep things up-to-date.


balsama:
Awesome. I did find at least one value that will vary depending on the environment and an additional UUID that will change:

1. In system.date.yml `timezone: default: VALUE`
1. In embed.button.media_browser.yml `icon_uuid: VALUE`

And it looks like `dependencies: content: file: VALUE` is also the UUID of the icon in embed.button.media_browser.yml.

So all of system.date.yml should probably be excluded. And there are probably several other system.*.yml files that could vary. I'm not sure if blacklisting them is a good approach or what.

PR #370, opened 2017-05-20
Bugfix: Resolves issue in which specified target bundle's source field file extensions are not restricting.

amcgowanca:
## Problem

File extensions specified on the media bundle's source field are not adhered too and do not restrict users from uploading files of other media bundle type's file extensions.

For example, an Image field which has a source field specifying an Image with file extensions set as `png, jpeg, jpg, gif` are not restricting users from uploading additional file extensions configured on _other_ media bundle's source fields that may not be of type image - an example would be a Document bundle, allowing for `pdf, txt, doc, docx` file extensions. These file types _could_ be uploaded within the Image media bundle as well.

### DIFF
--- a/modules/lightning_features/lightning_core/src/BundleEntityStorage.php
+++ b/modules/lightning_features/lightning_core/src/BundleEntityStorage.php
@@ -64,4 +64,23 @@ class BundleEntityStorage extends ConfigEntityStorage {
     return parent::loadMultiple($ids);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function loadByProperties(array $values = [], $check_access = FALSE) {
+    // Build a query to fetch the entity IDs.
+    $entity_query = $this->getQuery();
+    $this->buildPropertyQuery($entity_query, $values);
+    if ($check_access) {
+      $result = array_filter($entity_query->execute(), [
+        $this->accessHandler,
+        'createAccess'
+      ]);
+    }
+    else {
+      $result = $entity_query->execute();
+    }
+    return $result ? parent::loadMultiple($result) : [];
+  }
+
 }
--- a/modules/lightning_features/lightning_core/src/BundleEntityStorage.php
+++ b/modules/lightning_features/lightning_core/src/BundleEntityStorage.php
@@ -71,15 +71,12 @@ class BundleEntityStorage extends ConfigEntityStorage {
     // Build a query to fetch the entity IDs.
     $entity_query = $this->getQuery();
     $this->buildPropertyQuery($entity_query, $values);
+    $result = $entity_query->execute();
+
     if ($check_access) {
-      $result = array_filter($entity_query->execute(), [
-        $this->accessHandler,
-        'createAccess'
-      ]);
+      $result = array_filter($result, [$this->accessHandler, 'createAccess']);
     }
-    else {
-      $result = $entity_query->execute();
-    }
+
     return $result ? parent::loadMultiple($result) : [];
   }
 
--- a/modules/lightning_features/lightning_core/src/BundleEntityStorage.php
+++ b/modules/lightning_features/lightning_core/src/BundleEntityStorage.php
@@ -57,7 +57,7 @@ class BundleEntityStorage extends ConfigEntityStorage {
   public function loadMultiple(array $ids = NULL, $check_access = FALSE) {
     if ($check_access) {
       $ids = array_filter(
-        $this->getQuery()->execute(),
+        $ids ?: $this->getQuery()->execute(),
         [$this->accessHandler, 'createAccess']
       );
     }
--- a/modules/lightning_features/lightning_core/src/BundleEntityStorage.php
+++ b/modules/lightning_features/lightning_core/src/BundleEntityStorage.php
@@ -64,20 +64,4 @@ class BundleEntityStorage extends ConfigEntityStorage {
     return parent::loadMultiple($ids);
   }
 
-  /**
-   * {@inheritdoc}
-   */
-  public function loadByProperties(array $values = [], $check_access = FALSE) {
-    // Build a query to fetch the entity IDs.
-    $entity_query = $this->getQuery();
-    $this->buildPropertyQuery($entity_query, $values);
-    $result = $entity_query->execute();
-
-    if ($check_access) {
-      $result = array_filter($result, [$this->accessHandler, 'createAccess']);
-    }
-
-    return $result ? parent::loadMultiple($result) : [];
-  }
-
 }
--- a/modules/lightning_features/lightning_media/src/MediaHelper.php
+++ b/modules/lightning_features/lightning_media/src/MediaHelper.php
@@ -62,7 +62,7 @@ class MediaHelper {
         // If the field is a FileItem or any of its descendants, we can consider
         // it a file field. This will automatically include things like image
         // fields, which extend file fields.
-        if (!empty($field) && is_a($field->getItemDefinition()->getClass(), FileItem::class, TRUE)) {
+        if (is_a($field->getItemDefinition()->getClass(), FileItem::class, TRUE)) {
           $extensions .= $field->getSetting('file_extensions') . ' ';
         }
       }
--- a/modules/lightning_features/lightning_media/src/MediaHelper.php
+++ b/modules/lightning_features/lightning_media/src/MediaHelper.php
@@ -36,21 +36,21 @@ class MediaHelper {
    * @param bool $check_access
    *   (optional) Whether to filter the bundles by create access for the current
    *   user. Defaults to FALSE.
-   * @param string[] $target_bundles
-   *   (optional) An array of target bundles to retrieve source field extensions
-   *   for. If omitted, all bundles are allowed.
+   * @param string[] $bundles
+   *   (optional) An array of bundle IDs from which to retrieve source field
+   *   extensions. If omitted, all available bundles are allowed.
    *
    * @return string[]
    *   The file extensions accepted by all available bundles.
    */
-  public function getFileExtensions($check_access = FALSE, array $target_bundles = []) {
+  public function getFileExtensions($check_access = FALSE, array $bundles = []) {
     $extensions = '';
 
     // Lightning Media overrides the media_bundle storage handler with a special
     // one that adds an optional second parameter to loadMultiple().
     $storage = $this->entityTypeManager
       ->getStorage('media_bundle');
-    $bundles = $storage->loadMultiple($target_bundles ?: NULL, $check_access);
+    $bundles = $storage->loadMultiple($bundles ?: NULL, $check_access);
 
     /** @var \Drupal\media_entity\MediaBundleInterface $bundle */
     foreach ($bundles as $bundle) {
@@ -79,6 +79,9 @@ class MediaHelper {
    * @param bool $check_access
    *   (optional) Whether to filter the bundles by create access for the current
    *   user. Defaults to TRUE.
+   * @param string[] $bundles
+   *   (optional) A set of media bundle IDs which might match the input. If
+   *   omitted, all available bundles are checked.
    *
    * @return \Drupal\media_entity\MediaBundleInterface
    *   A media bundle that can accept the input value.
@@ -86,12 +89,12 @@ class MediaHelper {
    * @throws \Drupal\lightning_media\Exception\IndeterminateBundleException if
    * no bundle can be matched to the input value.
    */
-  public function getBundleFromInput($value, $check_access = TRUE) {
+  public function getBundleFromInput($value, $check_access = TRUE, array $bundles = []) {
     // Lightning Media overrides the media_bundle storage handler with a special
     // one that adds an optional second parameter to loadMultiple().
     $bundles = $this->entityTypeManager
       ->getStorage('media_bundle')
-      ->loadMultiple(NULL, $check_access);
+      ->loadMultiple($bundles ?: NULL, $check_access);
 
     /** @var \Drupal\media_entity\MediaBundleInterface $bundle */
     foreach ($bundles as $bundle) {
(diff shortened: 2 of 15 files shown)


balsama:
Thanks for this. I need to spend some time confirming, but I think we were aware of this and, ultimately, the main entity form will not validate on save if you try to place a disallowed extension. But you're correct in that it will allow you to place it in the field. So it would be nice to notify the user sooner.

Should have some time before Wednesday. We'll also need a test for this before merging.

amcgowanca:
@balsama I will see if I can put together a test for this over the next few days.

phenaproxima:
I have to admit that I don't quite understand the problem being fixed by this PR. Our media upload thing is not *supposed* to filter by bundle, because it determines which bundle to use *after* you have uploaded a file. Indeed, the type of upload is what is used to determine the bundle. It seems to me that filtering by bundle in the first place would utterly short-circuit that.

It's true that you can then upload files that are not allowed by the field in question, but what's the problem there? They are allowed by some other field in the system (or they wouldn't be in the list of allowed extensions), so that does not strike me as a security risk. The file will be uploaded, nothing will use it, and core will eventually clean it out.

So I guess I'd just like to understand what the benefit of this would be...sorry if I'm being dense!

amcgowanca:
@phenaproxima No worries at all. Let me try and explain a bit further, though, I am sure I can run through a scenario quicker for explanations. @TravisCarden may be able to set something up.

But ...

1. Content type "ABC" and Content type "XYZ" exist with a Media reference field for specifically uploading Images. The Media bundle is configured with a single Image field with the following allowed extensions: `gif, jpeg, jpg, png`. The field settings on Content type ABC and Content type XYZ are limiting bundles to only the "Image" media bundle.
2. Additional media bundles exist for other types references, etc. For example, a Document media bundle with a file upload field with the allowed file extensions: `txt, pdf, doc, docx`.
3. When a user browses to `/node/add/abc` for creating a new node of type ABC, the form display widgets are configured to allow a user to upload a file. In this case, the field is to only allow extensions `gif, jpeg, jpg, png`. However, a user is prompted to upload a file allowing them to choose any. If this user chooses a PDF, the file is upload and no validations occur to restrict the type of field.
4. The content model and usages of the media field in this case are to be used for rendering images within the site. The problem is that a no actual image here exists and instead a document bundle then would exist.

This then leads into another problem, which I am not entirely sure is a problem or is meant to be implemented as such. When multiple bundles are created, as you described above, the system is to automatically determine the type of bundle. The problem exists when the field is configured to allow _only_ bundles of a certain type, the entity creation process happening creates a bundle of the wrong type which therefore cannot be referenced based on the restricted media bundles as per the field settings for the Media reference. Furthermore, a problem uncovered on one project this past week was that should multiple media bundles exist in which the allowed file types are `jpeg, jpg, png, gif` - the first bundle allowed of this type is what is created instead of the target bundle type within the field settings.

balsama:
I believe that there is an interim validation error when trying to save the content form between steps 3 & 4. So you would never actually get to step 4. A problem still exists in that the user isn't warned when they upload and place the file.

amcgowanca:
@balsama , You are correct. Technically the user would not be able to save the node itself as the field validation constraint is imposed due to target bundle not being correct for referencing. So, does it not make sense to cause the validation error prior to a node save? A content author most likely is unaware of the following error message is referring to in the sense of what had happened given that they are presented with what appears to be a successful file upload:

> This entity (media: 526) cannot be referenced.

phenaproxima:
That makes a lot more sense, thanks @amcgowanca! As a UX issue, I agree that this needs to be properly addressed. I will review the code soon...

modules/lightning_features/lightning_core/src/BundleEntityStorage.php
```diff
@@ -64,4 +64,23 @@ public function loadMultiple(array $ids = NULL, $check_access = FALSE) {
...
+    if ($check_access) {
+      $result = array_filter($entity_query->execute(), [
+        $this->accessHandler,
+        'createAccess'
+      ]);
+    }
+    else {
+      $result = $entity_query->execute();
+    }
+    return $result ? parent::loadMultiple($result) : [];
```
phenaproxima: I'd rather do the if ($check_access) and array_filter() call before returning, rather than two separate calls to $entity_query->execute().

modules/lightning_features/lightning_media/src/MediaHelper.php
```diff
@@ -37,17 +37,26 @@ public function __construct(EntityTypeManagerInterface $entity_type_manager) {
    *   (optional) Whether to filter the bundles by create access for the current
    *   user. Defaults to FALSE.
    *
+   * @param array $target_bundles
```
phenaproxima: Nit -- there should not be a blank line above this parameter. Also, the type hint should be string[].

modules/lightning_features/lightning_media/src/MediaHelper.php
```diff
@@ -37,17 +37,26 @@ public function __construct(EntityTypeManagerInterface $entity_type_manager) {
...
+   *   (optional) An array of target bundles to retrieve source field extensions
+   *   for. Defaults to empty array.
```
phenaproxima: Can we change this to "If omitted, all bundles are allowed"?

modules/lightning_features/lightning_media/src/MediaHelper.php
```diff
@@ -37,17 +37,26 @@ public function __construct(EntityTypeManagerInterface $entity_type_manager) {
...
 
     // Lightning Media overrides the media_bundle storage handler with a special
     // one that adds an optional second parameter to loadMultiple().
-    $bundles = $this->entityTypeManager
-      ->getStorage('media_bundle')
-      ->loadMultiple(NULL, $check_access);
+    $storage = $this->entityTypeManager
+      ->getStorage('media_bundle');
+    if (!empty($target_bundles)) {
+      $bundles = $storage->loadByProperties(['id' => $target_bundles], $check_access);
```
phenaproxima: I don't understand why loadByProperties() is called here. Why not just do $storage->loadMultiple()? Something like  $storage->loadMultiple($target_bundles ?: NULL) would allow us to knock a few lines out here.

modules/lightning_features/lightning_media/src/Plugin/EntityBrowser/Widget/FileUpload.php
```diff
@@ -65,9 +65,12 @@ public function getForm(array &$original_form, FormStateInterface $form_state, a
     // it as the first validator, allowing it to accept only file extensions
     // associated with existing media bundles.
     if (empty($validators['file_validate_extensions'])) {
+      $entity_browser_info = $form_state->get('entity_browser');
```
phenaproxima: I think $form_state->get() will take an additional argument (the default value). To prevent potential weirdness, let's do $form_state->get('entity_browser', []) so that we always get an array back.

phenaproxima (changes requested):

balsama:
> does it not make sense to cause the validation error prior to a node save

Oh no, it definitely does. Just wanted to make sure we were all on the same page. It's currently terrible UX.

modules/lightning_features/lightning_media/src/MediaHelper.php
amcgowanca: @phenaproxima I think the problem that I faced when testing was that when using `loadMultiple()`, the `$target_bundles` values were *not* being correctly returned as aprt of the query _unless_ specified via `id` property. I will retest locally.

amcgowanca: @phenaproxima Just realizing this as I seemed to have overlooked it previously .... The reason `loadMultiple()` cannot be used is due to the actual method's implementation to begin with:

```
  public function loadMultiple(array $ids = NULL, $check_access = FALSE) {
    if ($check_access) {
      $ids = array_filter(
        $this->getQuery()->execute(),
        [$this->accessHandler, 'createAccess']
      );
    }
    return parent::loadMultiple($ids);
  }
```

You'll notice here that the `$ids` are being overwritten from the "get all" query that is happening in `$ids = array_filter($this->getQuery()->execute() ...`

This would be the reason in which I implemented the `loadByProperties` to mimic / workaround this issue to begin with.

Thoughts?

I'd assume that this actually would result in unexpected behaviour for any Lightning based project that is using the storage handler for interacting with the media bundles...

amcgowanca: @balsama , @phenaproxima  (@TravisCarden, @lahoosascoots)

Proposed implementation of the fix for the `loadMultiple()`:

```
  public function loadMultiple(array $ids = NULL, $check_access = FALSE) {
    if ($check_access) {
      $ids = $ids ?: $this->getQuery()->execute();
      $ids = array_filter($ids, [$this->accessHandler, 'createAccess']);
    }
    return parent::loadMultiple($ids);
  }
```

Thoughts?

modules/lightning_features/lightning_media/src/Plugin/EntityBrowser/Widget/FileUpload.php
amcgowanca: @phenaproxima To my knowledge, `FormStateInterface::get()` only accepts a single argument - `$property`. See: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Form%21FormStateInterface.php/function/FormStateInterface%3A%3Aget/8.3.x

amcgowanca: @phenaproxima Proposed refactor:

```
      $target_bundles = [];
      if (!empty($entity_browser_info['widget_context']['target_bundles'])) {
        if (is_array($entity_browser_info['widget_context']['target_bundles'])) {
          $target_bundles = $entity_browser_info['widget_context']['target_bundles'];
        }
      }
```

This way we can handle the case of `$entity_browser_info['widget_context']['target_bundles']` is not being an array.

modules/lightning_features/lightning_media/src/MediaHelper.php
```diff
@@ -62,7 +62,7 @@ public function getFileExtensions($check_access = FALSE, array $target_bundles =
         // If the field is a FileItem or any of its descendants, we can consider
         // it a file field. This will automatically include things like image
         // fields, which extend file fields.
-        if (is_a($field->getItemDefinition()->getClass(), FileItem::class, TRUE)) {
+        if (!empty($field) && is_a($field->getItemDefinition()->getClass(), FileItem::class, TRUE)) {
```
amcgowanca: @phenaproxima , @balsama -

It is possible under the following scenario in which the `$field` _could_ be empty and therefore would result in a fatal error due to treating `$field` as always being an object.

*Scenario:*

1. An administrator (or a user with permissions) _changes_ the field definition of a Media bundle by deleting the source field. For example, on the Image media bundle - the "image" field which is set as the source is deleted.
2. An administrator (or a user with permissions) adds a _new_ image field (not that which was previously set and now removed).
3. A content creator (someone with content creation access for creating new media by uploading a new file on the form widget (e.g. Entity Browser & File Upload) during the creation of a new content.

modules/lightning_features/lightning_media/src/MediaHelper.php
```diff
@@ -36,10 +36,9 @@ public function __construct(EntityTypeManagerInterface $entity_type_manager) {
    * @param bool $check_access
    *   (optional) Whether to filter the bundles by create access for the current
    *   user. Defaults to FALSE.
-   *
-   * @param array $target_bundles
+   * @param string[] $target_bundles
    *   (optional) An array of target bundles to retrieve source field extensions
-   *   for. Defaults to empty array.
+   *   for. If omitted, all bundles are allowed.
    *
```
amcgowanca: @phenaproxima I believe this is the change you were referring to?

modules/lightning_features/lightning_core/src/BundleEntityStorage.php
```diff
@@ -57,7 +57,7 @@ public static function createInstance(ContainerInterface $container, EntityTypeI
   public function loadMultiple(array $ids = NULL, $check_access = FALSE) {
     if ($check_access) {
       $ids = array_filter(
-        $this->getQuery()->execute(),
+        $ids ?: $this->getQuery()->execute(),
```
amcgowanca, line 60: @phenaproxima , @balsama -

As mentioned [here](https://github.com/acquia/lightning/pull/370#discussion_r118573023), the specified `$ids` were _always_ being overwritten when `$check_access` was `TRUE`.

modules/lightning_features/lightning_media/src/MediaHelper.php
```diff
@@ -50,12 +50,7 @@ public function getFileExtensions($check_access = FALSE, array $target_bundles =
     // one that adds an optional second parameter to loadMultiple().
     $storage = $this->entityTypeManager
       ->getStorage('media_bundle');
-    if (!empty($target_bundles)) {
-      $bundles = $storage->loadByProperties(['id' => $target_bundles], $check_access);
-    }
-    else {
-      $bundles = $storage->loadMultiple(NULL, $check_access);
-    }
+    $bundles = $storage->loadMultiple($target_bundles ?: NULL, $check_access);
```
amcgowanca: @phenaproxima , @balsama -

Only possible to with commit: [d96ffe1](https://github.com/acquia/lightning/pull/370/commits/d96ffe18d72442c942327c60ee6f8d79ae4518fe).

phenaproxima:
In discussion with @amcgowanca and @TravisCarden, I have agreed to write a regression test for this before we merge.

modules/lightning_features/lightning_core/src/BundleEntityStorage.php
phenaproxima: Good catch. Approved!

phenaproxima:
I wrote a regression test! It works for me on my localhost and seems to confirm both the problem, and the fix. Can you apply this patch and merge it into this PR? At that point, if it passes Travis CI and Pipelines (and it should), I think we can merge this.
[370-regression.txt](https://github.com/acquia/lightning/files/1038985/370-regression.txt)

phenaproxima (approved):

balsama:
This looks and works great. I also confirmed that this fixed the problem described in #374.

My only nit - and I think we went over this the last time we attempted to address this @phenaproxima - is that while the error message's timing and wording is vastly improved, the escaped HTML is super annoying. It could also benefit from some Element::oxford() love.

If these can't be addressed, can we please open up a followup issue (on D.O)? We're certain to get issues in our queue if not.

Thanks for the great issues and collaboration to fix!

phenaproxima:
Follow-up issue opened: https://www.drupal.org/node/2882718

PR #375, opened 2017-06-01
Bugfix: Resolves PHP notices when the prefix element is undefined.

amcgowanca:


phenaproxima:
Nice little fix. @amcgowanca, if you will add a comment explaining what you're fixing -- just a sentence or two, for posterity -- I think this is OK to merge.

phenaproxima:
Although, one thing does give me pause here: we're extending Entity Browser's widget, which means that we should be able to rely on their render array having a certain structure, and having certain keys. So now I'm not 100% sure this is something we *should* be fixing. At the very least, we should be aware of the conditions under which the prefix is not set, rather than just working around the possibility of it not being set.

amcgowanca:
@phenaproxima -

Ha, my apologies for not adding a comment. I was pushing up various PRs all morning and this one had slipped through the crack for a comment. I am going to re-test so I can provide exact scenario and circumstances where I ran into the issue this morning. I'll mark this as closed if it is in fact not relevant from what I uncover when I take another look. Thanks.

danepowell:
We started seeing these errors after upgrading to Lightning 2.1.4, presumably because of this commit: https://github.com/acquia/lightning/commit/7998fccef29b5c1a7829d5d76f6037c753df16a2

However they've been very difficult to reproduce consistently. The only thing I know for sure is that they appear on the node_add form for content types containing media reference / entity browser fields.

balsama:
@danepowell Can you confirm that you have patch 10 from 2877751 applied to Entity Browser?

EntityReferenceBrowserWidget::displayCurrentSelection populates `current`. That patch adds the last line in the snippet below to displayCurrentSelection - so I can't see any way `['current']['#prefix']` could _not_ be populated if that patch is applied.

```
    return [
      '#theme_wrappers' => ['container'],
      '#attributes' => ['class' => ['entities-list']],
      '#prefix' => '<p>' . $this->getCardinalityMessage($entities) . '</p>',
...
```

danepowell:
Yep that patch is definitely applied.

amcgowanca:
@balsama , @phenaproxima , @danepowell :

image

This is the stack trace from the PHP notice that is printed when I receive the issue. Just encountered it again this morning. I should also add that this is happening on both the Field UI (when the default value/widget is rendered) screen _and_ entity edit forms in which the field is being used.

PR #384, opened 2017-06-15
Behat - Added getEntities method.

modcab:
So we can retrieve created entities during the scenario and act on them.


phenaproxima:
We have ceased to support Behat for testing, so this is probably outdated. I'm closing it out -- thanks!

PR #389, opened 2017-06-20
Pinned to a slightly older version of Drupal Extension.

### LINKED ISSUE #385: Latest version of behat and drupal/extension cause fatal
Updating drupal/drupal-extension to 3.3.0 loosens its constraint on behat/behat which means it will update to 3.3.1. After updating, the `Creating media assets from within the media browser using embed codes` fails with the following:

```
@BeforeScenario # PanelsSubContext::gatherContexts()
Fatal error: Call to undefined method Behat\Behat\Context\Environment\UninitializedContextEnvironment::getContexts()
```

Which apparently comes from DrupalSubContextBase::getContext. What's weird is that I don't see UninitializedContextEnvironment ever containing a getContexts method. So I'm not sure what's up but I wanted to document it.

You can see test failure here: https://travis-ci.org/balsama/lightning/builds/243794888 or on lightning-project which doesn't have a lock file and thus updated already.

phenaproxima:
This fixes #385 by pinning to the 3.2.x line of Drupal Extension.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "phpunit/phpunit": "~4.8",
         "symfony/css-selector": "~2.8",
         "drush/drush": "8.1.11",
-        "drupal/drupal-extension": "^3.2",
+        "drupal/drupal-extension": "~3.2.0",
         "behat/behat": "^3.0",
         "drupal/coder": "~8.2.0",
         "phing/phing": "^2.14",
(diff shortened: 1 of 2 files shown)


balsama:
The lock file here doesn't reflect the constraints in the composer.json file. Can you just run:

```
composer update drupal/drupal-extension behat/behat
```

This also indicates that a more recent version of drupal-extension and behat _will_ technically work since the above downgraded behat v3.3.1 => v3.1.0 and drupal-extension v3.3.0 => v3.2.3.

phenaproxima:
I ran that command, but virtually nothing changed in composer.lock. And certainly nothing that references either drupal/drupal-extension or behat/behat. Not sure there's anything else I need to do here.

balsama:
hm. possible that I was mistaken and hadn't run install with your lock file I guess. Thanks.

PR #390, opened 2017-06-20
Changing Behat web driver to headless Chrome.

grasmash:


doriancmore:
@grasmash you need to configure the extension too https://gitlab.com/DMore/behat-chrome-extension#usage

https://github.com/acquia/lightning/blob/6c712bcbce30570c89a52ed731fdd9ab6c554b1c/src/Command/BehatInitCommand.php#L57

phenaproxima:
This is fixed in #579 :)

PR #396, opened 2017-06-27
Move install-state Behat tests to PHPUnit

phenaproxima:
A couple of our Behat tests don't actually test functionality, but only verify the state of various permissions and other install-time configuration. Now that we have a functional PHPUnit test that passes, these tests should not be done in Behat. This will shave a little time off our Travis CI runs.

### DIFF
--- a/tests/features/media/roles.feature
+++ b/tests/features/media/roles.feature
@@ -1,24 +1,6 @@
 @lightning @core @media @api
 Feature: Responsibility-based user roles for creating and managing media assets
 
-  @beta5 @80dd53a1
-  Scenario: Media-related user roles should exist
-    Given I am logged in as a user with the "administer permissions" permission
-    When I visit "/admin/people/roles"
-    Then I should see "Media Creator"
-    And I should see "Media Manager"
-
-  @beta5 @68d467b4
-  Scenario: Content creators have access to the rich_text input format and entity browsers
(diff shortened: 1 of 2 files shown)


balsama:
Just restarted 4/5 tests that failed on the "Users are notified that an internal view mode is internal" scenario (see [tracking spreadsheet](https://docs.google.com/spreadsheets/d/1Sb4KTZgVwPalLnsYk9pBblCjRsW2I1NBwsj0XuZgOtk/edit#gid=0)).

I don't see how those failures could be related to this PR, but 4/5 is remarkable. Let's see what happens on the restart.

balsama:
Hm. Same failures: https://travis-ci.org/acquia/lightning/builds/247747217

phenaproxima:
It passed on Pipelines, but failed on Travis, for reasons that have got to have absolutely nothing to do with the actual test being run? Something is wonky here. I'd say random fails, but the failures seem deterministic.

It's quite possible that there is a caching issue; "test-cleanup" has probably been used as a branch name by me before. I'll try clearing out all Travis-side caches.

balsama:
4/5 failed again. And always 2.1.3 that passes. Interesting. And, as you noted, pipelines passed.

Anyway, it's consistent enough that we should be able to reproduce outside of the tests tomorrow.

phenaproxima:
Closing in favor of #416.

PR #397, opened 2017-06-28
Make it possible to choose an image style and alt text at embed time

phenaproxima:
This PR addresses https://www.drupal.org/node/2784699.

As of this PR, you can select an image style, as well as set per-embed alt and title text, when you embed an image into a CKEditor instance. This means that there is no more need to have a view mode for each image style. It's not drag-handle resizing, which would be the best option, but it's not bad either.


lightning.install
```diff
@@ -457,7 +458,8 @@ function lightning_update_8001() {
   // Install the body field on the basic block_content type.
   $field = FieldConfig::load('block_content.basic.body');
   if (empty($field)) {
-    lightning_core_create_config('field_config', 'block_content.basic.body', 'lightning');
+    $config = Config::forModule('lightning');
```
phenaproxima, line 461: This is a bit of gardening, unrelated to the new functionality. lightning_core_create_config() has been gone for a long time now, so this code was fatally broken.

modules/lightning_features/lightning_core/modules/lightning_test/lightning_test.install
```diff
@@ -81,21 +81,6 @@ function lightning_test_install() {
       ],
       'region' => 'content',
     ])
-    ->setComponent('field_z_image', [
```
phenaproxima, line 84: All of this was moved into the new upload_bundles_test module, which is a new test-only module that was split out so as not to cause a conflict which prevented the CKEditor media browser from being properly testable when lightning_test is turned on.

modules/lightning_features/lightning_core/src/ConfigHelper.php
```diff
@@ -128,7 +128,7 @@ public function delete($id) {
       if (Unicode::strpos($id, $prefix) === 0) {
         $entity = $this->getEntity(
           $entity_type,
-          Unicode::substr($id, Unicode::strlen($prefix) + 1)
```
phenaproxima, line 131: This turned out to have always been broken, just never tested until now.

modules/lightning_features/lightning_media/config/install/filter.format.rich_text.yml
```diff
@@ -46,7 +44,7 @@ filters:
     status: true
     weight: -50
     settings:
-      allowed_html: '<a href hreflang> <em> <strong> <cite> <blockquote cite> <code> <ul type> <ol start type> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <s> <sup> <sub> <img src alt data-entity-type data-entity-uuid data-align data-caption> <table> <caption> <tbody> <thead> <tfoot> <th> <td> <tr> <hr> <p> <br> <h1> <pre> <drupal-entity data-*>'
```
phenaproxima, line 49: This change allows alt and title text to be overridden in a particular embed. See https://www.drupal.org/node/2834112

modules/lightning_features/lightning_media/lightning_media.install
```diff
@@ -270,9 +270,6 @@ function lightning_media_update_8015() {
 function lightning_media_update_dependencies() {
   return [
     'lightning_media' => [
-      8004 => [
```
phenaproxima, line 273: This is just gardening. Lightning update 8002 no longer does anything.

modules/upload_bundles_test
```diff
@@ -0,0 +1 @@
+../tests/modules/upload_bundles_test
```
phenaproxima, line 1: This is a symlink that allows upload_bundles_test to be turned on during testing. Modules are not discoverable except in certain places...go figure.

src/LightningExtension/Context/UtilityContext.behat.inc
```diff
@@ -27,33 +29,6 @@ class UtilityContext extends DrupalSubContextBase {
...
-   * @param int $count
-   *   The number of elements expected to match the selector.
-   *
-   * @throws ExpectationException
-   *   If the number of elements that match the selector is not the expected
-   *   number.
-   *
-   * @Then :count element(s) should match :selector
-   */
-  public function assertSelectorMatch($selector, $count) {
```
phenaproxima, line 43: I removed this because it turns out that the Mink extension already has a step definition that does the same thing. So this counts as gardening.

tests/features/bootstrap/lightning.behat.inc
```diff
@@ -19,28 +19,6 @@ use Drupal\DrupalExtension\Context\DrupalSubContextBase;
 class LightningSubContext extends DrupalSubContextBase {
 
   /**
-   * Installs the Lightning Test module before a @test_module scenario.
-   *
-   * @BeforeScenario @test_module
-   */
-  public function installTestModule() {
```
phenaproxima, line 26: This is replaced by TestModuleContext.

tests/features/media/browser_embed_code.feature
```diff
@@ -1,7 +1,7 @@
 @lightning @media @api @javascript @errors
 Feature: Creating media assets from within the media browser using embed codes
 
-  @video @twitter @instagram @test_module @2c43f38c
+  @video @twitter @instagram @with-module:lightning_test @2c43f38c
```
phenaproxima, line 4: New syntax! Now we can create smaller test modules for specific scenarios that need such scaffolding, rather than a gigantic monolith.

balsama:
Should we consider hiding the "Display as" select box? Under what circumstances would a user want to embed an image or use Media Image for something other than an image?

If we _don't_ hide it, I think we need either better labels and succinct description, or canonical documentation about what the heck it does. Drupal has a bad habit of putting options like this in interfaces without explaining them.

balsama:
When I embed an image using "Embed", save, and then edit the node, the "Display as" option for that image is set to "Media Image".

balsama:
Possible follow up:
Hide the select option and provide a way to toggle it on in config.

balsama (changes requested):
This looks and works great. I do think that we should address the labeling issue I raised in the comments. Here's my suggestion:

* Hide the select box by default. It will still behave the same - automatically defaulting to Media Image for images and Embedded for all other bundles
* Allow users to unhide that select box through an exposed configuration option
* Update the manual update steps to account for sites which might have view-mode assigned image styles and want to continue to use that workflow.

phenaproxima:
Okay, I have added all of that, plus a little bit more test coverage. There is now a Media tab at the Lightning settings page, and it has one checkbox which determines whether "Display as" will be shown. In a completely new installation of Lightning, it will be unchecked because the new default behavior is, obviously, to hide "Display as". However, in an updated installation, it will be checked, so as to preserve the previous behavior and be as non-disruptive as possible.

PR #400, opened 2017-06-29
Add UUIDs to field storage configs

### LINKED ISSUE #387: Managing Lightning configuration requires running entity-updates after site install
The ideal way to manage configuration for a site using Lightning would be something like this:

1. Install site initially using the Lightning profile (or subprofile)
2. Make project-specific modifications to site (adding custom modules, configuration, etc...)
3. Exported site configuration to disk (`drush cex`)
4. On subsequent installations, use the `--config-dir` option with `drush site-install` to import custom configuration added in step 2.

The problem is that currently, after that last step, and after cron has been run, you will have * Mismatched entity and/or field definitions* errors on your Status Report page. You will have to run `drush entity-updates` in order for Drupal to completely sync the configuration.

This is because Lightning creates fields at install time, and Drupal assigns these fields a random UUID. Then when you import the site configuration for those fields (from a previous install), Drupal needs to change the UUID. Unfortunately, this triggers an upstream bug, whereby [Drupal cannot change UUIDs during a config import](https://www.drupal.org/node/2876795), and instead requires cron and entity-updates to be run. This is a really bad developer experience for many reasons, including the fact that entity-updates is a pretty hacky and dangerous command that shouldn't be used on any regular basis.

I'm wondering if there's any way the Lightning team could help get traction on that upstream core issue, to improve the DX for Lightning as well.

Note that this has also been documented in this BLT issue: https://github.com/acquia/blt/issues/1297#issuecomment-300030478

balsama:
Fixes #387 from Lightning's perspective. Core still provides two field storage config definitions that don't include UUIDs (field.storage.block_content.body.yml and field.storage.node.body.yml). This adds UUIDs to all of Lightning's field storage config.

### DIFF
--- a/modules/lightning_features/lightning_core/config/install/field.storage.node.field_meta_tags.yml
+++ b/modules/lightning_features/lightning_core/config/install/field.storage.node.field_meta_tags.yml
@@ -1,3 +1,4 @@
+uuid: 8fac44a0-73ae-4055-b75c-1181b79f6608
 langcode: en
 status: true
 dependencies:
--- a/modules/lightning_features/lightning_layout/config/install/field.storage.node.panelizer.yml
+++ b/modules/lightning_features/lightning_layout/config/install/field.storage.node.panelizer.yml
@@ -1,3 +1,4 @@
+uuid: 69a4c29f-f3b3-4ebe-ac8f-b8f05651f528
 langcode: en
 status: true
 dependencies:
--- a/modules/lightning_features/lightning_media/config/install/field.storage.media.embed_code.yml
+++ b/modules/lightning_features/lightning_media/config/install/field.storage.media.embed_code.yml
@@ -1,3 +1,4 @@
+uuid: 62bf66dd-0ded-43f1-8a6d-21049dcf611f
 langcode: en
 status: true
 dependencies:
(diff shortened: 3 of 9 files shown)


danepowell (approved):

phenaproxima (changes requested):
I think we should probably do this for *all* of Lightning's bundled config, not just fields. Consistency matters. Additionally, we will need to fix our Phing clean-config target so that it leaves UUIDs intact, as well as update everything in tests/config (i.e., re-export with UUIDs). So this PR needs to be a whole hell of a lot bigger than 9 files changed! :)

balsama:
Just talked this over and agree with @phenaproxima. Nothing too challenging about what he's recommending, just tedious.

I'd also really like to get some clarity on why core doesn't provide UUIDs on the config it provides. There might be a perfectly good reason why we _shouldn't_ be doing this. (Or not, who knows. But I'd hate to have to revert)

danepowell:
I agree that it's suspicious that core doesn't do this. I suspect it's more out of principle (i.e. "default configuration should be as generic as possible") than real specific practical reasons.

balsama:
Closing this PR for now while we continue to discuss in #387.

Issue #403, opened 2017-07-13
Lightning Core's hook_entity_load implementation causing necessary slowness


billdennen:
We've noticed some slowness after upgrading from Lightning 1 to 2.15. According to New Relic, the slowness is the result of Lightning Core's hook_entity_load.

balsama:
It looks like that code is vestigial and can be removed.

balsama:
Just to make sure we weren't missing something, I manually tested path aliases, including with pathauto and manual overrides, with this hook and without it. I couldn't find any unexpected behavior or difference between when it was and wan not present.

Tests are running here: https://travis-ci.org/balsama/lightning/builds/253637252 If they pass, I'll file a PR.

PR #406, opened 2017-07-14
Remove vestigial hook_entity_load

balsama:
See #403.

I don't think that this hook could have caused significant slowness, but it appears to be unused and should be removed.


phenaproxima (approved):
<3

PR #412, opened 2017-07-26
Introduce a component for API-first site architectures

phenaproxima:
This PR addresses https://www.drupal.org/node/2896267.

At a high level -- this introduces a new top-level Lightning component, called Content API (the machine name is lightning_api). It's a loose, dependency-free stub of a module around which we bring in JSON API, OpenAPI, and Simple OAuth, with little additional configuration except for providing a friendly link to the auto-generated API docs, and generating an initial key pair for OAuth, if possible. All of this is adapted from/based on the code found in Reservoir, specifically its install routines. The purpose is to provide a very basic server-side framework for building decoupled apps that use Lightning as a backend.

Our goal is not to compete with something as feature-complete and focused as Contenta, but simply to allow people to interact with the entities in Drupal using a standard JSON-based API. We expect more things will happen in this space, but right now the goal is just to choose our basic tool set and lay a good foundation.


phenaproxima:
@wimleers, @mattgrill, @tedbow, @aweingarten: When reviewing, please ignore any changes in the tests/config directory. Those are part of the way we ensure that our config stays up-to-date, and don't need any actual review from you.

balsama:
Consumers will likely run into cross-domain policy issues when setting up their application. We should consider documenting how to add CORS response headers via the Drupal's services.yml.

Contenta has a nice, concise, section about this in their README: https://github.com/contentacms/contenta_jsonapi#cors

.travis.yml
```diff
@@ -80,7 +80,7 @@ install:
   - phing recall -Dversion=$VERSION -Ddb.database=drupal -Ddb.user=lightning -Ddb.password=lightning -Ddb.host=127.0.0.1
   - cd docroot
   # Execute database updates, if there are any.
-  - drush updatedb --yes
+  - drupal update:execute
```
balsama, line 83: If this is really necessary (that is, drush doesn't handle caches properly) then we should update UPDATE.md too and explain why you *must* use DrupalConsole.

phenaproxima: It is necessary to get the tests to pass, but can we update the documentation in another PR?

mattgrill (approved):
👍🎉 The Drupal'ly parts of this, I don't have any feedback on. Overall I think it's great to add the modules and give Lightning a robust API.

.travis.yml
balsama: At a minimum, we should update bullet 4 here: https://github.com/acquia/lightning/blob/8.x-2.x/UPDATE.md#tarball

A detailed description might not be needed, but we shouldn't give bad instructions.

phenaproxima: Drush works for the updates, as long as you clear the cache afterwards to resolve inconsistencies. Is it OK if I update that bullet so that it mentions an explicit cache clear after the updates are done?

balsama:
* During install, we warn about generating a new keypair before deploying. We should provide documentation around how to do that.
* We need a new section under "What Lightning Does" in the README file for Content API.

UPDATE.md
```diff
@@ -65,6 +67,10 @@ Beta 2, then from Beta 2 to Beta 3, in that order.
   actively using Page Manager, you must add it to your project as an explicit
   dependency in order to continue to using it. Otherwise, you should uninstall
   it before updating to the next version of Lightning.
+* Lightning now supports exposing all Drupal entities as JSON, in the standard
+  JSON API format. To enable this feature, install the OpenAPI, OpenAPI Redoc
+  Documentation, and JSON API modules. Then install the Content API module from
+  the Lightning package.
```
tedbow: Not really what is going on in this description. Why aren't these modules just dependencies of Content API. Why does the user have turn on these module independently

tedbow: Probably need to understand this before I can really comment on any of the reset.

phenaproxima: The reason they are not dependencies is because there was no need for them to be. They are all installed by the Lightning profile, but Content API is not actually doing anything which outright *requires* any of them. To keep things relatively cherry-pickable (i.e., for PS), I prefer to avoid defining hard dependencies when they're not needed.

balsama: I understand why it was set up this way - it makes sense and it will save us from dependency problems/complaints later on.

Still, I feel like people will want to just turn this on to experiment. And this is a barrier.

Could we add a subcomponent of Content API that does what Content API currently does and make that and the contrib modules a dependency of Content API? That way, there's a single module that turnkey adds the functionality - but no direct dependencies between the functional modules.

lightning.info.yml
```diff
@@ -36,10 +36,7 @@ dependencies:
   - views
   - views_ui
   - diff
-  - metatag
-  - jsonapi
-  - openapi_redoc
-  - simple_oauth_extras
+  - metatagΩ
```
balsama: well there's your problem :)

.idea/inspectionProfiles/profiles_settings.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
```
balsama: Whoops. Looks like an phpstorm file got added to this commit somehow. Remove please?

balsama:
7bb19e10192e7c9d89e2150a08686ecce94c887d 👍

PR #419, opened 2017-08-01
Make shipped configuration optional

phenaproxima:
As we have discovered in the process of building acquia/headless-lightning, Lightning doesn't really facilitate easy extensibility by shipping all of its packaged config in the required config/install directory. All of it, except for user.role.administrator, can be optional -- which will free up sub-profiles to exclude Bartik, Seven, Block, and Block Content.

Addresses https://www.drupal.org/node/2899017


phenaproxima:
Committed in abfe5cfa75ef6fec0167425fc46dbd1ceb15b5a4.

PR #421, opened 2017-08-02
Provide inline links to API documentation for bundle config entities

phenaproxima:
The OpenAPI ReDoc module generates API documentation for all bundle config entities (such as node types, media types, etc.) A nice little thing that Lightning API could do -- and Headless Lightning could take advantage of -- is to provide an inline "View API documentation" link for all such entities. This PR does exactly that, revealed in the operations drop-button of each bundle config entity listed in a standard table.

I didn't bother with tests because this is very low-risk and doesn't alter user data in any way. But I can add one if needed.

### DIFF
--- a/modules/lightning_features/lightning_api/lightning_api.module
+++ b/modules/lightning_features/lightning_api/lightning_api.module
@@ -1,5 +1,7 @@
 <?php
 
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Url;
 use Drupal\lightning_api\OAuthHelper;
 
 /**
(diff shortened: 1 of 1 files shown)


dgpalmer:
Is there a particular reason why this is being added to the main lightning distribution and not a feature that can be turned on or off like lightning layout? We are using lightning, but have many features turned off, and when I ran a composer update today, I saw these 2 new modules come through, the jsonapi and openapi. Is there a reason why this is being used when there is a rest api provided by Drupal 8?

I am just concerned cause we are in the process of building our first mobile app integrated with Drupal 8 for user generated content and this is adding something that could potentially conflict with it.

It seems to me there are a lot of assumptions being made and enforced with lightning. It would be nice if this stuff was modular and could be turned off like what you guys did with lightning_layout.

Nothing is broken but I just wanted to express my concerns.

phenaproxima:
There are a number of reasons why we're including this functionality, but you reopening this PR raises a good point to me -- it should be possible to remove the API documentation links. I will open a new pull request to add a config switch behind which those can be hidden.

As for the new decoupled/API-first functionality, it can most definitely be turned off: simply uninstall the Content API module (from the Lightning package) and its dependencies, which include JSON API and OpenAPI. As with Lightning Layout, or just about any other part of Lightning, you don't need to use things you don't need or want. In fact, if you have an existing Lightning installation, none of the API stuff will be turned on by the automatic updates. We have been moving towards making things quite modular indeed, and dependencies across Lightning components are already pretty loose.

dgpalmer:
Thanks Adam! Yah, we have a sub profile of lightning and have excluded the things we do not want, which works great.

phenaproxima:
Cool. If you're using a sub-profile, you should have no problem -- just exclude the lightning_api module and you will not get any of the new and shiny :)

dgpalmer:
Ah, perfect. Thank you!

PR #422, opened 2017-08-03
Add a grid layout for administrative link lists in Seven

phenaproxima:
This originates from Headless Lightning, but I'm porting it here so we can re-use it in vanilla Lightning. It adds a class to Seven which allows us to take lists of administrative links -- like you see at /admin/structure or /admin/config -- and make them into lovely flexbox-ey grids. With optional support for adding icons to the links.


balsama:
This looks good. Will merge when the tests go green.

PR #425, opened 2017-08-08
Support bulk upload of media assets

phenaproxima:
This is a one-two punch of win that addresses https://www.drupal.org/node/2672038 and https://www.drupal.org/node/2870739 simultaneously.

It takes advantage of the improvements in Composer 1.5 to bring in the Dropzone JavaScript library, and the DropzoneJS Drupal module, to provide a bulk upload form in Lightning. This is the reference implementation of my preferred standard way to import JavaScript libraries into Lightning -- via Asset Packagist.


composer.json
```diff
@@ -152,7 +161,10 @@
         "drupal/search_api": "^1.0",
         "drupal/jsonapi": "^1.1",
         "drupal/openapi": "1.0-alpha1",
-        "drupal/simple_oauth": "^2.0@RC"
+        "drupal/simple_oauth": "^2.0@RC",
+        "oomphinc/composer-installers-extender": "^1.1",
+        "drupal/dropzonejs": "^1.0@alpha",
```
balsama, line 166: This project is alpha stability which means we must pin to a release.

balsama: We should get rid of the superfluous @ stability tags too. If we're going to assume the scaffold project has the proper installer paths, we can assume it has stability set to dev too.

balsama: Would be nice if we could require a minimum global version of composer (1.5.0) here (like you can for php). But I don't think there is a way to do that. If you require composer/composer it just puts it in vendor and provides a bin.

balsama (changes requested):
Excited that we might get this in soon. We need PRs against BLT, DF, and Lightning Project that add

* The required additional installer paths
* The asset packagist repository
* Additional installer types

And those PRs need to be merged before we can merge this. Is there anything else that a scaffold project needs to change?

PR #426, opened 2017-08-08
Run composer validate in non-interactive mode

balsama:


phenaproxima:
Looks good to me. Merge when green :)

PR #430, opened 2017-08-14
Support bulk upload of media assets (#425)

balsama:
* Updated dependencies and added a service provider to Lightning API which disables Metatag's normalizers.

* Added bulk upload module.

* Added DropzoneJS module and library.

* Restored bulk upload test and update documentation.

* Accepting composer validate prompt automatically.

* Granted media_creator access to Dropzone by default.


composer.json
```diff
@@ -152,7 +161,10 @@
         "drupal/search_api": "^1.0",
         "drupal/jsonapi": "^1.1",
         "drupal/openapi": "1.0-alpha1",
-        "drupal/simple_oauth": "^2.0@RC"
+        "drupal/simple_oauth": "^2.0@RC",
+        "oomphinc/composer-installers-extender": "^1.1",
+        "drupal/dropzonejs": "^1.0@alpha",
```
balsama: Need to pin to a release since this is alpha stability

balsama: Let's get rid of the superfluous stability tags

phenaproxima: Done.

balsama:
Downstream PRs:

* Lightning: acquia/lightning-project#41
* DF: acquia/df#58
* BLT: acquia/blt#1923

phenaproxima:
@DevinCarlson, @mortenson, @briancwald, and @brantwynn: This PR is going to add DropzoneJS, and the accompanying library (latest version) to Lightning. I know that acquia/df is already bringing that in, pinned to v4.2.0. This PR doesn't pin to a specific version of the Dropzone library, so by default it pulls in the current 5.x version. Will this break Demo Framework? Or should I file a PR to make DF rely on Lightning's bundled Dropzone library, once we merge this?

DevinCarlson:
@phenaproxima Don't worry about it. After the new Lightning release, I'll drop the dependency from DF and defer to Lightning; we always want to work with whatever modules or libraries you include.

As for dropzone 5.x specifically, I did some basic testing and it looks like everything is working correctly. We usually pin to specific library releases for convenience and because we weren't using something like composer asset plugin to help manage them.

phenaproxima:
This is blocked by #431.

balsama:
This is RTBC from me. When https://github.com/acquia/blt/pull/1923 is merged, this can be committed.

PR #431, opened 2017-08-15
Add support for Asset Packagist

phenaproxima:
This PR addresses https://www.drupal.org/node/2870739.

It adds Asset Packagist to our composer.json as a repository, and switches us from composer/installers to oomphinc/composer-installers-extender so that NPM and Bower packages from Asset Packagist can be moved into their proper place in docroot.

It also includes a Composer script which can be used to make the appropriate adjustments to the composer.json of a project that is consuming Lightning (e.g., a Lightning Project instance). All of this is documented at (currently unpublished): http://lightning.acquia.com/blog/round-your-front-end-javascript-libraries-composer

This blocks #430, and is blocked by the following pull requests:

acquia/lightning-project#41 (merged)
acquia/df#58 (merged)
acquia/blt#1923


composer.json
```diff
@@ -53,15 +57,22 @@
...
+        ],
         "installer-paths": {
             "docroot/core": [
                 "type:drupal-core"
             ],
             "docroot/libraries/{$name}": [
-                "type:drupal-library"
+                "type:drupal-library",
+                "type:bower-asset",
+                "type:npm-asset"
```
mortenson, line 75: @phenaproxima @balsama Are `installer-paths` inherited, or will every Lightning user have to copy+paste these to their `composer.json` file?

phenaproxima: I think every Lightning user will need to copy and paste those lines, just as they would with anything else in the extras.installer-paths section. This is documented in the blog post I mentioned in the OP.

mortenson: So for all Lightning users, this will be a manual update they have to perform to get the bulk upload functionality working?

balsama: > every Lightning user will need to copy and paste those lines

Or run the included [Composer script](https://github.com/acquia/lightning/blob/8.x-2.x/src/Composer/AssetPackagist.php):

```
composer run enable-asset-packagist
```

phenaproxima: Technically yes, but in practice no, because we're providing a script to make the required modifications to composer.json. And that will only affect existing users of Lightning; it'll just work for new users.

PR #439, opened 2017-08-24
Support image cropping

phenaproxima:
At long, long last: this addresses https://www.drupal.org/node/2690423.

This adds Crop API and Image Widget Crop to Lightning as an *optional* quasi-dependency of Lightning Media Image. If Image Widget Crop is installed, four crop types are created: freeform, portrait, landscape, and square (thanks @starshaped for suggesting those). Each crop type, and every crop type created thereafter, also gets an accompanying image style which is used to display that particular crop.

In new installs, or after manual updates in existing installs, the Image media type will by default use the cropping widget in both its normal form and in the media browser (which is also used by the image browser). All available crop types will be allowed, but that can be overridden per form display.

@mirodietiker had suggested using Focal Point as our cropping widget, but I decided to stick with Image Widget Crop because it is a more "traditional" form of cropping than Focal Point is, and I figure Lightning shouldn't venture too far off the reservation in terms of expected out-of-the-box CMS behavior. There is nothing in this PR to prevent people from switching to Focal Point if they want to, though.


composer.json
```diff
@@ -169,7 +173,10 @@
         "drupal/simple_oauth": "2.0.0-rc3",
         "oomphinc/composer-installers-extender": "^1.1",
         "drupal/dropzonejs": "1.0.0-alpha7",
-        "bower-asset/dropzone": "^5.1"
+        "bower-asset/dropzone": "^5.1",
+        "drupal/crop": "1.2.0",
```
phenaproxima, line 179: I had to add Crop API as an explicit dependency here in order to patch it. Once the patch is committed, we can remove this dependency, since Image Widget Crop already depends on Crop API.

drupal-org.make
```diff
@@ -72,12 +77,17 @@ projects[token][version] = 1.0
...
 projects[video_embed_field][version] = 1.5
 projects[views_infinite_scroll][type] = module
-projects[views_infinite_scroll][version] = 1.3
+projects[views_infinite_scroll][version] = 1.5
 projects[workbench_moderation][type] = module
 projects[workbench_moderation][patch][] = https://www.drupal.org/files/issues/2668006-2.patch
 projects[workbench_moderation][patch][] = https://www.drupal.org/files/issues/2847078-6.patch
 projects[workbench_moderation][patch][] = https://www.drupal.org/files/issues/workbench_moderation-quickedit-support-2749503-6.patch
 projects[workbench_moderation][version] = 1.2
+libraries[cropper][type] = library
```
phenaproxima, line 86: This library is required by Image Widget Crop and it's already on drupal.org's library packaging whitelist.

lightning.info.yml
```diff
@@ -36,6 +35,7 @@ dependencies:
   - views
   - views_ui
   - diff
+  - image_widget_crop
```
phenaproxima, line 38: Image Widget Crop has a hard dependency on Image.

modules/lightning_features/lightning_core/lightning_core.module
```diff
@@ -17,6 +17,13 @@ use Drupal\lightning_core\OverrideHelper as Override;
 use Drupal\lightning_core\Plugin\views\filter\Bundle;
 
 /**
+ * Implements hook_local_tasks_alter().
+ */
+function lightning_core_local_tasks_alter(array &$local_tasks) {
+  $local_tasks['entity.node.edit_form']['options']['attributes']['rel'] = 'edit-form';
```
phenaproxima, line 23: Gardening. This helps with testing when we need Behat to click the Edit tab on a node page, but we don't care what the tab's text is.

modules/lightning_features/lightning_core/modules/lightning_dev/lightning_dev.install
```diff
@@ -54,6 +55,24 @@ function lightning_dev_install() {
...
+  $image_styles = array_map(
+    function ($crop_type) {
+      return 'crop_' . $crop_type;
+    },
+    \Drupal::entityQuery('crop_type')->execute()
+  );
+  $image_styles = ImageStyle::loadMultiple($image_styles);
+
+  /** @var ImageStyle $image_style */
+  foreach ($image_styles as $image_style) {
```
phenaproxima, line 68: This is necessary because image effects get a new UUID when they're added to an image style. That would cause our exported config to change with every new install, which would be no bueno.

modules/lightning_features/lightning_media/config/schema/lightning_media.schema.yml
```diff
@@ -35,5 +35,13 @@ entity_browser.browser.widget.file_upload:
       type: boolean
       label: 'Return underlying file entity'
 
-field.widget.settings.image_immutable:
-  type: field.widget.settings.image_image
+field.widget.third_party.lightning_media:
+  type: mapping
```
phenaproxima, line 39: Turns out, you can inject third-party settings into field widget and formatter configuration! I had no idea you could do this, and it's super-useful, but the settings must have a config schema defined. Thus, this.

modules/lightning_features/lightning_media/lightning_media.install
```diff
@@ -278,16 +279,24 @@ function lightning_media_update_8016() {
...
-    ],
-  ];
+function lightning_media_update_8017() {
+  \Drupal::service('plugin.manager.field.widget')->clearCachedDefinitions();
+
+  /** @var \Drupal\Core\Entity\Display\EntityFormDisplayInterface $form_display */
+  foreach (EntityFormDisplay::loadMultiple() as $form_display) {
+    foreach ($form_display->getComponents() as $key => $component) {
+      if ($component['type'] == 'image_immutable') {
+        $component['third_party_settings']['lightning_media'] = [
```
phenaproxima, line 291: We no longer need the image_immutable widget -- we can achieve the same effect in a more streamlined fashion by using third-party widget settings. This migrates from image_immutable to those.

modules/lightning_features/lightning_media/lightning_media.module
```diff
@@ -25,15 +29,63 @@ use Drupal\media_entity\MediaInterface;
 use Drupal\media_entity\MediaBundleInterface;
 
 /**
- * Implements hook_field_widget_info_alter().
+ * Implements hook_field_widget_third_party_settings_form().
  */
-function lightning_media_field_widget_info_alter(array &$info) {
-  if (isset($info['entity_browser_entity_reference'])) {
-    Override::pluginClass($info['entity_browser_entity_reference'], EntityReferenceBrowserWidget::class);
+function lightning_media_field_widget_third_party_settings_form(WidgetInterface $widget) {
```
phenaproxima, line 34: All of these changes constitute a major refactor of how we were customizing the behavior of image widgets. This is a *far* more stable way to achieve what we were previously doing with plugin hijacking.

modules/lightning_features/lightning_media/lightning_media.module
```diff
@@ -345,12 +397,8 @@ function lightning_media_ajax_render_alter(array &$data) {
...
     }
 
-    // Gingerly check if the AJAX command class exists, so that we don't need
-    // to establish a hard dependency on CKEditor.
-    $command_class = '\Drupal\ckeditor\Ajax\AddStyleSheetCommand';
-    if ($style_sheets && class_exists($command_class)) {
-      /** @var \Drupal\ckeditor\Ajax\AddStyleSheetCommand $command */
-      $command = new $command_class($query->get('editor'), $style_sheets);
+    if ($style_sheets) {
+      $command = new AddStyleSheetCommand($query->get('editor'), $style_sheets);
```
phenaproxima, line 401: This is gardening. Lightning Media has a hard dependency on the CKEditor module, so there was no need to gingerly check anything.

modules/lightning_features/lightning_media/modules/lightning_media_image/lightning_media_image.module
```diff
@@ -8,10 +8,50 @@
...
 use Drupal\lightning_media_image\Plugin\MediaEntity\Type\Image;
 
 /**
+ * Implements hook_field_widget_info_alter().
+ */
+function lightning_media_image_field_widget_info_alter(array &$info) {
+  if (isset($info['image_widget_crop'])) {
+    // We cannot import the class with a use statement, because it will blow up
+    // if Image Widget Crop is not installed.
+    Override::pluginClass($info['image_widget_crop'], \Drupal\lightning_media_image\Plugin\Field\FieldWidget\ImageCropWidget::class);
```
phenaproxima, line 23: To get the behavior I wanted, I had no choice but to hijack the cropping widget plugin.

modules/lightning_features/lightning_media/modules/lightning_media_image/src/Plugin/Field/FieldWidget/ImageCropWidget.php
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getSetting($key) {
+    $value = parent::getSetting($key);
+
+    // If no crop types are chosen, allow all of them.
+    if ($key == 'crop_list' && empty($value)) {
```
phenaproxima, line 46: The default behavior of this widget is to only allow crop types which are explicitly enabled for that specific form display. Syncing that with new or deleted crop types would be a complicated pain in the ass from both a DX and UX perspective, so I overrode it. Site builders will still be able to explicitly allow only certain crop types for a particular form display.

modules/lightning_features/lightning_media/src/Plugin/Field/FieldWidget/EntityReferenceBrowserWidget.php
```diff
@@ -1,41 +0,0 @@
...
-namespace Drupal\lightning_media\Plugin\Field\FieldWidget;
-
-use Drupal\Core\Field\FieldItemListInterface;
-use Drupal\Core\Form\FormStateInterface;
-use Drupal\entity_browser\Plugin\Field\FieldWidget\EntityReferenceBrowserWidget as BaseEntityReferenceBrowserWidget;
-
-/**
- * Cosmetic enhancements for the Entity Browser entity reference widget.
- */
-class EntityReferenceBrowserWidget extends BaseEntityReferenceBrowserWidget {
```
phenaproxima, line 12: This has been replaced by a hook implementation in lightning_media.module, so that we don't need to bother with plugin hijacking.

src/LightningExtension/AwaitTrait.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Acquia\LightningExtension;
+
+use Behat\Mink\Exception\UnsupportedDriverActionException;
+use Drupal\DrupalExtension\Context\MinkContext;
+
+trait AwaitTrait {
```
phenaproxima, line 8: This is gardening of the tests. It's a lot nicer than having to pull in MinkContext and call iWaitForAjaxToFinish().

src/LightningExtension/Context/UtilityContext.behat.inc
```diff
@@ -91,29 +138,6 @@ class UtilityContext extends DrupalSubContextBase {
...
   /**
-   * Finds a collapsible details element by its summary text.
-   *
-   * @param string $summary_text
-   *   The summary text.
-   *
-   * @return \Behat\Mink\Element\NodeElement|null
-   *   The details element, or NULL if it was not found.
-   */
-  public function findCollapsible($summary_text) {
```
phenaproxima, line 102: This is replaced by DetailsTrait::assertDetails().

tests/config/lightning_api.settings.yml
```diff
@@ -1,2 +1,3 @@
 entity_json: true
+bundle_docs: false
```
phenaproxima, line 2: Whoops. We should have had this in a previous PR, but I guess we forgot. No harm done.

tests/config/views.view.content.yml
```diff
@@ -329,16 +329,31 @@ display:
           hide_empty: false
           empty_zero: false
           hide_alter_empty: true
-          destination: true
-          plugin_id: entity_operations
-        moderation_state:
-          id: moderation_state
+          click_sort_column: target_id
```
phenaproxima, line 332: This is the result of a previous PR which moved the operations field. Again, I think we just forgot to merge updated config then. NBD.

tests/features/media/image.feature
```diff
@@ -16,7 +16,22 @@ Feature: Image media assets
     Then I should be visiting a media entity
     And I should see "Foobaz"
 
-  @image @javascript @b23435a5
+  @13eacffd
+  Scenario: Cropping should be allowed when creating an image
```
phenaproxima, line 20: It'd be really difficult to test the actual cropping, and that's squarely in Image Widget Crop's wheelhouse anyway. I figure it's good enough to ensure that the crop widget is showing up where and when we expect it to.

balsama:
This is excellent and very thorough. Thank you.

I have some questions strictly related to the UI. Some are almost definitely not related to Lightning's implementation, but we should investigate:

1. I feel like four options out of the box is a little overwhelming. I think the largest use case is an arbitrary crop. Can we only enable "Freeform" out of the box? We can leave them defined, but I think it would be best to enable one type on the Media Image image field at least.
2. It's not clear to me what the settings at /admin/config/media/crop-widget do. Are they just defaults for new fields that use crop. I tried to modify the settings there thinking it would edit Media: Image before I realized that those settings are on the field itself. Perhaps we could open an issue for  Image Widget Crop to clarify what that form does.
3. /admin/config/media/crop drop-buttons have a link to manage moderation. Is this the case for all entity lists like this? Can we disable it for non-moderated entity types?
4. The "This crop affects multiple images" warning message seems to be displayed on every image, regardless of whether it's being used or not.

Again, this is great. Thanks so much for this.

phenaproxima:
1. Fixed. The Image media type will use Freeform only, by default.

2. I took a quick check and TBH it's not clear to me either. I imagine they'd be treated as the default settings for new fields, but the crop widget itself does not appear to pull default settings from configuration or anything. Maybe it's a vestigial part of Image Widget Crop? In any event, it's an issue with Image Widget Crop, not Lightning. So I'm open to creating an issue (or including a patch) to either document or remove those settings.

3. I suspect this is an oversight in Workbench Moderation itself. Crop types (config entities) are "bundles" of crops (which are content entities), and I think Workbench Moderation is picking up on that pattern and assuming that moderation is a thing that crop types may want. We can suppress that behavior for crop types specifically, but I would rather whitelist than blacklist. I think we could open a follow-up PR to add an administrative UI to Lightning Workflow which will allow you to say which bundle entity types can have moderation applied. That would solve this problem permanently and generically.

4. I don't really understand that warning, TBH. There is a setting on the field which can disable that warning, though it's not clear in which contexts. I could set that as the default. What do you think?

balsama:
1. Cool. Looks like the tests and manual update steps need to be updated (manual update steps documentation and lightning_dev)
2. Let's open that issue when we get a chance, but not block on it
3. Let's open that issue against Lighting before merging this
4. Yes, let's turn off that setting. I'd rather not see that warning at all if possible.

phenaproxima:
1. Done.
2. Okay. That sounds like a plan to me.
3. Opened #447.
4. The setting in question now defaults to FALSE.

balsama:
Awesome. Let's let this build out to cloud one more time for a quick manual test. Will be here when complete: http://lightningnightlyfpce6h2qzh.devcloud.acquia-sites.com/

balsama:
Shoot. Disabling everything but Freeform causes another UX issue. The others still exist as crop types and each has an associated view mode so the user is still presented with NINE options when embedding. Three of which are identical. (And it's probably not immediately obvious to the user that they could just go enable them on the field)

I like the idea of providing sane defaults, but to keep this as un-intrusive as possible while still providing much needed functionality, let's just ice the config for everything but freeform entirely.

balsama:
Hm. Still see a "Crop thumbnail" image style that I didn't see before this PR (in addition to the "Cropped: Freeform" style that I expected to be there. Fresh install: http://lightningnightlyfpce6h2qzh.devcloud.acquia-sites.com/admin/config/media/image-styles

UPDATE.md
```diff
@@ -57,6 +57,36 @@ to Beta 3, you will need to follow the instructions for updating from Beta 1 to
...
 ## 2.1.7 to 2.1.8
+* Lightning now ships with support for image cropping, using the Image Widget
+  Crop module. To use it for the Image media bundle (the default behavior in
+  new Lightning sites), do the following:
+  * Install the Image Widget Crop module.
+  * Visit *Structure > Media bundles*. For the Image media bundle, choose
+    "Manage form display".
+  * If the Image field is enabled, set its widget type to "ImageWidget crop",
+    and configure it like so:
+    * Only "Freeform" should be selected for "Crop Type".
```
balsama: For "Preview image style", select "Crop thumbnail"

balsama:
Committed in 0665e16e17213a651b5387129035cf94233ff0e6 with the following CI tweaks:

* Bumped Travis CI php version to 7.0.22 because symfony/browser-kit requires >=7.0.8 if using php 7.0
* Pinned Drupal Console to 1.0.1 because 1.0.2 returns a non-zero exit code for update:execute if there are no pending updates

PR #440, opened 2017-08-25
Show real OAuth errors when generating a key pair.

mortenson:
If any errors are encountered when generating key pairs with Lightning's OAuthHelper, a generic error message ("An internal error occurred") is always thrown. This makes troubleshooting failures extremely difficult.

OpenSSL makes its most recent error available with the function `openssl_error_string()`(see http://php.net/manual/en/function.openssl-error-string.php), and Lightning should be presenting this message instead of the generic one, when possible.


phenaproxima:
I like this. Will merge when it goes green.

phenaproxima:
Thank you, @mortenson!

PR #449, opened 2017-09-06
Ensure Entity Browser filters are functional after SA-CORE-2017-004

### LINKED ISSUE #441: Entity Browser Views AJAX is broken in latest release
As of Drupal 8.3.7, all Entity Browser Views have broken exposed filters, pagers and sorts. There is an Entity Browser issue to address this, and I've confirmed that the most recent patch resolves the issue: https://www.drupal.org/node/2902831

Replication steps: open the Lightning Media browser and try to use the exposed filters, or click "Load more" at the bottom of the list. You should see a 403.

balsama:
Fixes #441

### DIFF
--- a/src/LightningExtension/Context/EntityBrowserContext.behat.inc
+++ b/src/LightningExtension/Context/EntityBrowserContext.behat.inc
@@ -3,6 +3,7 @@
 namespace Acquia\LightningExtension\Context;
 
 use Behat\Mink\Exception\ElementNotFoundException;
+use Behat\Mink\Exception\ExpectationException;
 use Drupal\Component\Utility\Html;
 use Drupal\DrupalExtension\Context\DrupalSubContextBase;
 use Drupal\DrupalExtension\Context\MinkContext;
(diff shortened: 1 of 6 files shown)


src/LightningExtension/Context/EntityBrowserContext.behat.inc
```diff
@@ -106,4 +107,40 @@ class EntityBrowserContext extends DrupalSubContextBase {
...
+   *   The number of media items that should be visible.
+   * @param string $browser_id
+   *   (optional) The entity browser ID.
+   *
+   * @throws ExpectationException
+   *   If actual number of in the entity browser does not match the expected.
+   *
+   * @Then I should see :n items in the entity browser
+   */
+  public function assertNumberVisible(int $n, $browser_id = NULL) {
```
phenaproxima: We can't type hint $n as an int unless we raise our minimum PHP version to 7 :)

src/LightningExtension/Context/EntityBrowserContext.behat.inc
```diff
@@ -106,4 +107,40 @@ class EntityBrowserContext extends DrupalSubContextBase {
...
+   *
+   * @param int $n
+   *   The number of media items that should be visible.
+   * @param string $browser_id
+   *   (optional) The entity browser ID.
+   *
+   * @throws ExpectationException
+   *   If actual number of in the entity browser does not match the expected.
+   *
+   * @Then I should see :n items in the entity browser
```
phenaproxima: Let's support single-count checks by making this: "I should see :n item(s) in the entity browser"

src/LightningExtension/Context/EntityBrowserContext.behat.inc
```diff
@@ -106,4 +107,40 @@ class EntityBrowserContext extends DrupalSubContextBase {
...
+    else {
+      $selector = 'form[data-entity-browser-uuid]';
+    }
+
+    /** @var \Behat\Mink\Element\NodeElement[] $items */
+    $items = $this
+      ->assertSession()
+      ->elementExists('css', $selector)
+      ->findAll('css', '[data-selectable]');
+    $actual = count($items);
```
phenaproxima: All of the preceding logic should be moved into a protected method and re-used, since it's copied directly from selectItem() and we should probably not be repeating ourselves :)

phenaproxima (changes requested):
Looks good, just a few changes.

balsama:
Indeed, indeed, indeed! 9eade10

phenaproxima (approved):

Issue #443, opened 2017-08-31
Generated OAuth keys have incorrect permissions when codebase is built with provided tarball


balsama:
If you use the [tarball provided on the Release page](https://github.com/acquia/lightning/releases/tag/2.1.7) to build your codebase (instead of the Composer [scaffold project](https://github.com/acquia/lightning-project#get-started)) the keys generated during the installation of Lightning API are set to 644 instead of 600 (or 660 which is also acceptable).

phenaproxima:
Fixed in #451.

PR #451, opened 2017-09-07
Lock down generated OAuth key permissions

phenaproxima:
I love the fact that this security-related PR addresses #443, which is the SSL port number. All this does is set 600 (very restrictive) permissions on the generated keys. Also added tests to assert this.


balsama:
I manually tested installing from a legacy build via the UI and did not receive a notice about the file perms. Looks good.

danepowell:
Note that Git can't actually store permissions other than the +x bit, which wreaks all kinds of havoc if you are trying to deploy keys with your Git repository. In other words you can generate keys with the correct permissions, but you can't put them in Git.

Issue #445, opened 2017-09-01
Determine workflow for OAuth keys


balsama:
Lightning API generates a pair of OAuth keys on install, saves them to a random directory inside the system temp dir, and then creates config to reference the new keys.

The problem is that the config that references the keys will be exported/imported in a normal workflow, but the keys the config references will not because they are outside of VCS root.

Possible solutions:
* Since we indicate that the keys are for development only, and should be changed before going into production, we could put the keys in a known location (like VCS root, one level up from docroot).
* Document implementing something like Config Split to exclude the keys config for all environments. Perhaps Lightning could actually configure that if Config Split is present?

balsama:
@damontgomery Did you come across this issue with the Reservior + BLT POC you built? Any suggestions on how to best handle it?

damontgomery:
There are two issues.

1. The tmp directory probably gets cleared regularly.
Solution: I store the keys in `/oath/private.key` if you want to commit them to the repo. Probably not the best long term solution, but OK for dev. Since it's probably not a good idea to spread the keys around, I recommend not creating keys for people.

2. The key directory needs to be absolute.
Solution: Store the keys in a relative directory and modify the OAuth module to use a relative path (I haven't tried this).

Solution: Don't export the OAuth configuration. Each site should be it's own client with it's own auth keys, so exporting doesn't make sense unless you use config split. Because the paths are absolute, also need different settings on local then on the server. One temporary solution (probably not a great long term solution either) is to upload the keys to the hosted server in a directory accessible to the web host that you can control permissions. You could add this to a deploy process like cloud hooks on Acquia hosting.

balsama:
I think a good workaround for now is:

1. Stop generating keys on install
2. Provide an option to automatically generate keys somewhere in the UI
3. The key generation UI should ask for an absolute path in which to store them. We should recommend the /home directory, since it's persistent and reliable on Acquia Cloud)
4. The /home directory _may_ be shared by multiple environments, but that is not guaranteed. Considering this, we should continue to give the keys unique name so that there are no conflicts between environments.
5. The user will be responsible for excluding the config for they from export/import since it will be different on different environments (using Config Split or otherwise).

balsama:
From @danepowell:

> One thing to be aware of for troubleshooting purposes: simple_oauth calls realpath on the key directory, which requires that PHP have execute permissions on every directory in the file path. This makes it somewhat brittle and can cause unexpected problems in some environments.

balsama:
[...]

PR #454, opened 2017-09-11
Manually generate OAuth key pairs

phenaproxima:
Addresses #445 by adding a dedicated form under the Lightning administrative area, at which the user can generate OAuth key pairs.


balsama:
This looks great. I just want to test cloud build and tarball manually. Will commit after that. Thanks.

balsama:
Hmm. I'm not sure if this a problem or not, but when I tested generation on cloud, the pubic key has 600 permissions instead of 400 immediately after generation. Then when I use the API, the private key also gets set to 600. While I don't see why they would need write permission, I can't think a reason that's a problem.

modules/lightning_features/lightning_api/lightning_api.module
```diff
@@ -2,35 +2,27 @@
...
-      $error = t('Could not generate a key pair because @reason. Until this is done, OAuth authentication will not work.', [
-        '@reason' => lcfirst($e->getMessage()),
-      ]);
-      drupal_set_message($error, 'error');
+    catch (RouteNotFoundException $e) {
+      \Drupal::service('router.builder')->rebuild();
     }
+
+    drupal_set_message(
+      t('You may wish to <a href=":generate_keys">generate a key pair</a> for OAuth authentication.', [
```
balsama: If installing through the UI, this message is displayed on the first page load and it ends up linking to "/core/install.php/admin/config/system/lightning/api/keys" as a result (note the install script in the path).

modules/lightning_features/lightning_api/src/Form/OAuthKeyForm.php
```diff
@@ -0,0 +1,310 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class OAuthKeyForm extends ConfigFormBase {
+
+  /**
+   * The UNIX permission bits to apply to the generated keys.
+   *
+   * @var integer
+   */
+  const PERMISSIONS = 0400;
```
balsama: Let's change this to 0600 since Simple OAuth ends up doing that anyway

modules/lightning_features/lightning_api/tests/src/Kernel/OAuthKeyFormTest.php
```diff
@@ -0,0 +1,46 @@
...
+      ->get('form_builder')
+      ->submitForm(OAuthKeyForm::class, $form_state);
+
+    $this->assertKey($dir . '/private.key');
+    $this->assertKey($dir . '/public.key');
+  }
+
+  private function assertKey($path) {
+    $this->assertFileExists($path);
+    $this->assertSame(0400, fileperms($path) & 0777);
```
balsama: And we'll need to change to 0600 here too.

PR #458, opened 2017-09-13
Make it easier to run Behat tests

phenaproxima:
Running our Behat tests is a bit of a pain in the butt at the moment. It requires me to memorize and type a long-ass command (and from the correct directory, too):

```
$ bin/behat --strict --stop-on-failure --config docroot/sites/default/files/behat.yml --tags=sometag path/to/my/test.feature
```

Ugh.

We can do better. I propose we include Robo as a dev dependency and write a command which can run our Behat tests more easily (it will also start Selenium up automatically in the background so we don't have to remember to do that too). With this command, from the top-level directory, this is how you'd run all Behat tests:

```
$ bin/robo test:behat
```

To run a specific feature, or directory of features:

```
# Tests a single feature, path relative to the tests/features dir
$ bin/robo test:behat -- api.feature

# Runs all tests in tests/features/media
$ bin/robo test:behat -- media

# Runs all tests with a specific tag (you can also exclude tags, just like normal)
$ bin/robo test:behat -- --tags=foobaz
```

Everything after the first -- will be passed directly to Behat, with minimal processing, so you can still use any Behat command-line switch you want.

Nice, right? This will make it less annoying for us to verify broken tests, just by making it easier to run them repeatedly.


.travis.yml
```diff
@@ -85,15 +85,13 @@ install:
...
-  - mkdir ../tmp
-  - selenium-server-standalone -port 4444 -log ../tmp/selenium2.log &
   - sleep 5
 
 script:
   - composer validate ../composer.json --no-check-all --ansi --no-interaction
   - phpunit --configuration ./core --group lightning
-  - behat --strict --stop-on-failure --config ./sites/default/files/behat.yml
+  - cd ..
+  - bin/robo test:behat
```
balsama, line 94: It's not immediately obvious that I can pass a tag as an argument. Also, can I pass more than one? What about excluding tags?

phenaproxima: Added:

```
$ bin/robo test:behat -- --tags=foo,~bar
```

Everything after the first -- is passed directly to Behat.

balsama:
Looks great. Let's add the documentation you provided here to the code itself.

PR #462, opened 2017-09-15
Introduce automatic configuration updates

phenaproxima:
Manual update steps are cool for documentation, but annoying to actually execute in a lot of cases. Generally, a simple script would suffice -- the only thing we've ever really wanted is to be able to ask the user to confirm actions that will modify their current config.

Well...meet this PR. This introduces a new Drupal Console command, update:lightning, which implements a lightweight, semver-based system for running configuration updates automatically. Modules can expose classes (one per version) containing annotated methods that perform the configuration updates. Console will confirm with the user before executing each update method.

We'll still have to document manual update steps, but this is a MUCH better way to run them, especially for users who want to just...do the updates, their existing config notwithstanding.


balsama:
Tested this pretty thoroughly. Love it.

My only question is that it seems I can run the `update:lightning` command several times and the command seems unaware that I have already run it. Which makes me wonder about the following scenario:

1. I run `update:lightning` for 2.1.8
2. I reorder my /admin/content view columns
3. I update to 2.1.9 and run the `update:lightning` again

Would the `update:lightning` command for 2.1.9 be unaware that I had already run the 2.1.8 updates and, therefore, do not want Lightning to update the /admin/content view again?

phenaproxima:
Hmmm...it is *definitely* supposed to remember that it has run the 2.1.8 updates. Otherwise, as you point out, it will try to run them again, which could blow stuff away. So if it's not remembering this stuff, that is absolutely a bug.

If I have time tomorrow (waiting to board my plane?) I'll try to write a unit test to confirm that the updated version is recorded in config, as it's supposed to be.

balsama:
I just double checked and it definitely allows me to run the updates more than once. I also confirmed that it resets any customizations I have done. Steps:

1. phing recall -Dversion=2.1.7
1. drupal update:execute
1. drupal update:lightning
1. Move Operations column in admin/content
1. Confirm Operations column is not last column
1.  drupal update:lightning
1. Opertaions column is back to last again

Unfortunately, I feel like we should block on this unless we can confirm that it's something easy to fix after release.

phenaproxima:
I agree that this blocks commit. I'll see what I can do today, or tomorrow, about fixing that.

phenaproxima:
Reproduced, and fixed!

balsama:
Excellent! Bon voyage!

Issue #459, opened 2017-09-14
Stop patching Drush and add uninstalling deps of parent profiles to known issues


balsama:
Currently, Drush won't allow you to uninstall dependencies of of parent profiles. There is a patch available but we don't actually need that patch for development on Lightning. The patch forces projects that use Lightning to use the same major version of Drush since the patch won't apply across version.

Let's drop the patch and add a reference to it in our Known Issues section. I would say we should drop Drush altogether, but we use it to interact with our nightly build deployed to cloud.

Issue #463, opened 2017-09-18
Update drush/drush to 9


naveenvalecha:
BLT allowed updating the drush to 9.0.0-beta4 See https://github.com/acquia/blt/issues/1985
https://github.com/acquia/blt/commit/6cdde6988e9811675d0d4565ccaf660fb32775cf

While updating the BLT to latest version, I tried to update the drush as well to the latest version, but it breaks because lightning stick with 8.1.13

balsama:
We actually want to have no opinion at all on which version of drush consumers use. See #459.

balsama:
Lightning no longer has an opinion about which version of Drush consumers use. See #464 f54afb52c802b00208aca5192478d01c68ff6db5

PR #464, opened 2017-09-18
Stop patching Drush and document inherited profile dependencies

balsama:
Fixes: #459 and #463


phenaproxima (approved):
Looks great. Cut and print!

Issue #479, opened 2017-10-04
"Publish" checkbox on node entities should be hidden


balsama:
The value of it is completely ignored and overridden by Lightning's "Moderation state" drop down. It's confusing and it should be removed from node bundles that have Moderation enabled.

balsama:
Fixed in #480

PR #480, opened 2017-10-05
Hide the publishing status checkbox from our default content types

phenaproxima:
Addresses #479.


balsama:
This is great. Thanks.

I think we should also cover new content types that have moderation enabled going forward, and also update existing content types (other than the two that Lightning provides).

Agree? Can we do that in this issue?

balsama:
If you have manually hidden the "Published status" checkbox (which I expect a lot of sites to do between now and when 2.2.1 comes out) the Console command still prompts you to hide it - which could be a little disconcerting to people. Ideally, I would like to adjust the logic such that users are only prompted to hide the checkbox on moderated content types that have their "Published status" checkbox visible in the form display.

If that's in anyway difficult to detect, we should just adjust the message:

> Do you want to ensure the "Publishing status" checkbox is hidden on the @node_type content type form?

PR #481, opened 2017-10-06
Require "Document" field on "Document" media bundle

TravisCarden:
Shouldn't the "Document" field on the "Document" media bundle be required? It hardly makes sense to create a media entity without a file, and all the other file fields on media bundles are required.


phenaproxima:
Yes, it should be. I think we should merge this. Not sure if we have to account for anything in the update path, though...

balsama:
Thanks @TravisCarden. We'll need to add this to the manual update steps in UPDATE.md and UpdateCommand. Which makes me think that we should document UpdateCommand since there's no way the submitter would have known to do that.

phenaproxima:
We'll need to merge this into acquia/lightning-media, but +1 to commit it.

phenaproxima:
This is now done in acquia/lightning_media, so I'm closing it out.

PR #488, opened 2017-10-19
Migrate to core Media module

phenaproxima:
Oh man.


balsama:
Since so much work was done on other branches forks since this PR, I opened #495. Once that goes green, we should merge that and close this.

balsama:
Merged as part of #495

PR #499, opened 2017-10-27
Fix hook_requirements issues

balsama:
The updates in 2.2.1 will fail to execute if hook_requirements is properly executed prior to the updates. The only reason it worked in the past was because drush had a bug where hook_requirements was *not* being run.

This PR addressed two separate issues with hook_requirements:
* The reset patch to media entity inadvertently removed the version from the info file - fixed in the reset patch
* Media Entity cached it's type plugins which caused media_entity_requirements to think modules like Media Entity Twitter were providing outdated plugins.


balsama:
Due to some problems with Robo 1.1.5 and our update scripts, Robo has been pinned to 1.1.4. And since Drush 9.0.0-beta8 requires Robo 1.1.5, we've also pinned that to beta7 (otherwise you get alpha1 which had looser dependencies).

phenaproxima:
Looks OK to me.

I'd like to know why Robo 1.1.5 breaks our update script, but that's not important just now. Right now, we need to stop the bleeding, so I'll merge this once it passes Travis CI.

PR #507, opened 2017-11-01
Install Composer dependences in docroot

phenaproxima:
This will make it easier to generate tarballs for use by Cloud. It doesn't affect our local build (as evidenced by the passing test suite on my branch), so I see no real reason not to do this.


balsama:
Turns out the classmap still has references to docroot even if vendor is inside it. Unfortunately this approach won't work because of that.

PR #515, opened 2017-11-06
Update image widget crop to core media

### LINKED ISSUE #514: 2.22 should have drupal/crop:^2.0
https://www.drupal.org/project/crop/releases/8.x-2.0-beta1 says:

> This is the first beta release with support for the core media entity. You should only update to this version if you are using core version.

This is part of the `drush media-entity-check-upgrade` output I got:

`The following modules depend on media entity and need to be either upgraded or uninstalled: crop_media_entity.`

The module mentioned is part of `drupal/crop`.

Additionally, since `drupal/image_widget_crop` depends (in its current form) on `drupal/crop:1.x`, it needs a patch (which I just created) included which takes its dependency to the correct level as well.

kekkis:
These changes should fix #514.

I tried to follow the style present elsewhere in the `composer.json` file. Feedback welcome, naturally.


kekkis:
Gah. I just realised it's not possible to use a _patch_ to amend a composer.json of a dependency. Instead we'll have to use the 2.x-dev#bcc23af.

Another realisation hit me reading through the Travis log: I had not included an updated composer.lock. I'll try to come up with that too.

drupal-org.make
```diff
@@ -35,7 +35,10 @@ projects[entity_embed][version] = 1.0-beta2
 projects[features][type] = module
 projects[features][version] = 3.5
 projects[image_widget_crop][type] = module
-projects[image_widget_crop][version] = 2.0
```
woprrr, line 38: Now that's unecessary ImageWidgetCrop support 1.x and 2.x branches of Crop API.

composer.json
```diff
@@ -191,7 +191,7 @@
         "drupal/dropzonejs": "1.0.0-alpha8",
         "bower-asset/dropzone": "^5.1",
         "drupal/crop": "^2.0-alpha1",
-        "drupal/image_widget_crop": "^2.0",
+        "drupal/image_widget_crop": "2.x-dev#bcc23af",
```
woprrr: Should use '^2.1' instead ?

composer.json
```diff
@@ -153,6 +153,10 @@
             "drupal/media_entity_instagram": {
                 "2917454 - Field formatter test is broken":
                 "https://www.drupal.org/files/issues/2917454-2.patch"
+            },
```
woprrr: This is merged in 2.1 now to be more flexible.

woprrr:
This Upgrade should use following patch to works with Media In core and avoid Fatal(s). https://www.drupal.org/node/2918441

balsama:
See #519

Issue #516, opened 2017-11-08
Lightning Image Widget Crop requirements


woprrr:
As we saw on the issue https://github.com/acquia/lightning/issues/514 it would be necessary to update the requirements of the module Image Widget Crop in order to be able to support the possibility to use crop API 1.x with media entity and to be able to update crop API in 2 .x without being blocked by the requirements of the Image Widget Crop module. During the last release 2.1.0 this compatibility is add and you can thus foresee an Update without modifying or patching Image Widget Crop.

To conclude if we use Media in core composer.json should be 
`        
"drupal/crop": "^2.0",
"drupal/image_widget_crop": "^2.1"
`
Or if we use Media Entity
`        
"drupal/crop": "^1.3",
"drupal/image_widget_crop": "^2.1"
`

balsama:
Fixed in 213afd204b5722c0e57cf23d72350c3b54edb37c

PR #517, opened 2017-11-08
Update ImageWidgetCrop module to support both branch of Crop API.

woprrr:
Follow-up of #516 This should fixe all problems from #515


balsama:
Thanks! I opened a new PR with the patch. Will fix there. #519

PR #518, opened 2017-11-08
Fix broken Media Entity update path

phenaproxima:
Lightning needs to include the patch from https://www.drupal.org/node/2921983, since it makes use of views with display media items.


balsama:
The 2.2.0 update path seems broken. It gets stuck on media_entity_update_8201:
https://travis-ci.org/acquia/lightning/jobs/299289427#L1200

PR #521, opened 2017-11-17
Temporarily patch core to make media name available on manage display and add update path to hide it.

balsama:
Added a test so that we can confirm functionality still works when it gets committed/we remove patch. See issue #2922289.

### DIFF
--- a/modules/lightning_features/lightning_media/src/Update/Update224.php
+++ b/modules/lightning_features/lightning_media/src/Update/Update224.php
@@ -36,6 +36,8 @@ final class Update224 implements ContainerInjectionInterface {
   protected $mediaDefinition;
 
   /**
+   * The entity type manager.
+   *
    * @var \Drupal\Core\Entity\EntityTypeManager
    */
   protected $entityTypeManager;
--- a/modules/lightning_features/lightning_media/src/Update/Update224.php
+++ b/modules/lightning_features/lightning_media/src/Update/Update224.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Drupal\lightning_media\Update;
+
+use Drupal\Console\Core\Style\DrupalStyle;
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\Core\Entity\EntityStorageInterface;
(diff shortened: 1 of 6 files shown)


modules/lightning_features/lightning_media/src/Update/Update224.php
```diff
@@ -0,0 +1,80 @@
...
+   * @var \Drupal\Core\Entity\EntityTypeInterface
+   */
+  protected $mediaDefinition;
+
+  /**
+   * The moderation information service.
+   *
+   * @var \Drupal\workbench_moderation\ModerationInformationInterface
+   */
+  protected $moderationInfo;
```
phenaproxima: We're not using this, so it should be removed.

modules/lightning_features/lightning_media/src/Update/Update224.php
```diff
@@ -0,0 +1,80 @@
...
+      $container->get('entity_type.manager')->getDefinition('media'),
+      $container->get('string_translation')
+    );
+  }
+
+  /**
+   * @update
+   */
+  public function hideNameField(DrupalStyle $io) {
+    \Drupal::entityTypeManager()->clearCachedDefinitions();
```
phenaproxima: The entity type manager should be injected.

modules/lightning_features/lightning_media/src/Update/Update224.php
```diff
@@ -0,0 +1,80 @@
...
+    /** @var MediaTypeInterface[] $media_types */
+    $media_types = $this->mediaTypeStorage->loadMultiple();
+    foreach ($media_types as $media_type) {
+      $question = (string) $this->t('Do you want to ensure the "Name" field is hidden on the Embedded view mode for the @media_type media type?', [
+        '@media_type' => $media_type->label(),
+      ]);
+
+      if ($io->confirm($question)) {
+        $display = entity_get_display('media', $media_type->id(), 'embedded');
```
phenaproxima: Nit: We don't need $display as a variable, we can just chain on the return value of entity_get_display().

phenaproxima (changes requested):
Looks good, just minor nits.

PR #524, opened 2017-11-29
Require composer patches 1.6.3.

danepowell:
Composer patches 1.6.3 was [just released](https://github.com/cweagans/composer-patches/releases/tag/1.6.3), which fixes [this pernicious bug](https://docs.acquia.com/article/fixing-failing-composer-patches-updating-gnu-patch), regardless of peoples' git/patch versions.

I think it would be good to require this new version. BLT is doing the same.

### DIFF
--- a/acquia-pipelines.yml
+++ b/acquia-pipelines.yml
@@ -17,6 +17,8 @@ events:
             # Add Composer's local bin directory to the PATH so that we will be
             # running our installed versions of Drush, PHPCS, Behat, etc.
             - export PATH="$HOME/.composer/vendor/bin:$SOURCE_DIR/bin:$PATH"
+            - git --version
+            - patch --version
             - composer validate --no-check-all --ansi --no-interaction
             - composer install
       - install:
--- a/acquia-pipelines.yml
+++ b/acquia-pipelines.yml
@@ -20,7 +20,7 @@ events:
             - git --version
             - patch --version
             - composer validate --no-check-all --ansi --no-interaction
-            - composer install
+            - composer install -vvv
       - install:
           type: script
           script:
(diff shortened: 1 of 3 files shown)


danepowell:
This is not encouraging... it looks like patches are silently failing to apply in Pipelines, although Travis is working fine.

This is doubly bad, because (a) this is using the supposedly fixed version of composer-patches, and (b) Pipelines is running patch 2.7.5, which should be immune to this problem regardless of the composer-patches fix.

The only other suspicious thing about Pipelines is its ancient Git version (2.1.4), but I just tested that locally and it works fine, patches are still applied.

So we're going to have to do some deep debugging here to figure out why patches aren't applying in Pipelines any more.

balsama:
Yeah, can confirm that at least the "2883813 - Move File/Image media type into Standard once Media is stable" patch is *not* applying even though the output from composer-patches indicates that it is.

The only obvious difference between Pipelines and Travis CI is that Travis is running PHP 7.0 and Pipelines is running 7.1.

danepowell:
Okay I confirmed that this is caused by the combination of Git 2.1.4 and Composer Patches 1.6.3

Upstream issue: https://github.com/cweagans/composer-patches/issues/174

mikemadison13:
Just to throw a bit more information here, I am on a project where we are still running PHP 5.6 across the board. This issue is NOT a problem locally, and it is NOT a problem on Travis, but it IS on Pipelines. Happy to share more information internally if it would be useful.

balsama:
As of cd52b8c3a6fdf8ae97d5ae626f527c6896bb0b50 we require composer-patches 1.6.4.

PR #530, opened 2017-12-05
Disable scheduled state changes on unmoderated content types

phenaproxima:
Addresses https://www.drupal.org/project/lightning/issues/2928429.


balsama:
Can we also add the following to the info file in this PR?

```
version: '8.x-2.24-dev'
package: Lightning
```

PR #537, opened 2017-12-11
Wait for Media Browser tab to load before interacting with it

balsama:


phenaproxima:
It would be preferable to wait for an element to appear, rather than a specific amount of time. Do we have a step for that?

balsama:
> It would be preferable to wait for an element to appear, rather than a specific amount of time. Do we have a step for that?

Not for a specific element or text, but Drupal Extension provides iWaitForAjaxToFinish() which should be better than the two seconds - which is based on nothing.

PR #539, opened 2017-12-12
Make ComponentDiscovery more flexible

balsama:
This PR allows Lightning ComponentDiscovery to work even if Lightning Components aren't inside the profile. So ComponentDiscovery will continue to work when Lightning Components are intermingled with other extensions.

It also adds a couple assertions that make sure other contrib and core modules aren't accidentally included.

### DIFF
--- a/src/ComponentDiscovery.php
+++ b/src/ComponentDiscovery.php
@@ -4,7 +4,6 @@ namespace Drupal\lightning;
 
 use Drupal\Core\Extension\Extension;
 use Drupal\Core\Extension\ExtensionDiscovery;
-use Drupal\field\Tests\reEnableModuleFieldTest;
 
 /**
  * Helper object to locate Lightning components and sub-components.
--- a/src/ComponentDiscovery.php
+++ b/src/ComponentDiscovery.php
@@ -11,6 +11,11 @@ use Drupal\Core\Extension\ExtensionDiscovery;
 class ComponentDiscovery {
 
   /**
+   * Prefix that Lightning components are expected to start with.
+   */
+  const COMPONENT_PREFIX = 'lightning_';
+
+  /**
    * The extension discovery iterator.
    *
    * @var \Drupal\Core\Extension\ExtensionDiscovery
@@ -70,7 +75,7 @@ class ComponentDiscovery {
    */
   public function getAll() {
     if (is_null($this->components)) {
-      $identifier = 'lightning_';
+      $identifier = self::COMPONENT_PREFIX;
 
       $filter = function (Extension $module) use ($identifier) {
         return strpos($module->getName(), $identifier) === 0;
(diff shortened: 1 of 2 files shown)


src/ComponentDiscovery.php
```diff
@@ -4,6 +4,7 @@
 
 use Drupal\Core\Extension\Extension;
 use Drupal\Core\Extension\ExtensionDiscovery;
+use Drupal\field\Tests\reEnableModuleFieldTest;
```
phenaproxima: Er -- this seems like it shouldn't be here...?

src/ComponentDiscovery.php
```diff
@@ -80,10 +71,10 @@ protected function getBaseComponentPath() {
    */
   public function getAll() {
     if (is_null($this->components)) {
-      $base_path = $this->getBaseComponentPath();
+      $identifier = 'lightning_';
```
phenaproxima: We use $identifier twice, so it should probably be a class constant. static::COMPONENT_PREFIX or something.

src/ComponentDiscovery.php
```diff
@@ -98,10 +89,17 @@ public function getAll() {
...
-      return dirname($module->getPath()) == $base_path;
+    $identifier = 'lightning_';
+
+    $filter = function (Extension $module) use ($identifier) {
+      // Assumes that:
+      // 1. Lightning sub-components are always in a sub-directory within the
+      //    main component.
+      // 2. The main component's directory starts with "lightning_".
+      // E.g.: "/lightning_core/modules/lightning_search".
+      $path = explode('/', $module->getPath());
```
phenaproxima: We should explode on DIRECTORY_SEPARATOR, not /.

src/ComponentDiscovery.php
```diff
@@ -98,10 +89,17 @@ public function getAll() {
...
+      $parent = $path[count($path)-3];
```
phenaproxima: What if $path has less than 3 elements already?

phenaproxima (changes requested):

src/ComponentDiscovery.php
balsama: Good question. If that's the case, the extension must be directly in the "/modules" directory which would mean it's definitely not a sub-component. In that case, $parent would end up getting set to `null` which is in turn used as the haystack in `strpos()` - which is used to determine if this is a sub-component - which is definitely is not! :)

Obviously not very self-documenting :/ agree this should probably be improved.

balsama:
@phenaproxima Any ideas on how to better handle `getMainComponents()`? Specifically the up-three-levels stuff? As it stands, it works reliably, but it's not very clear.

phenaproxima (approved):

PR #541, opened 2017-12-13
Make the moderation history view compatible with Content Moderation

phenaproxima:
Addresses https://www.drupal.org/project/lightning/issues/2930288.

### DIFF
--- a/modules/lightning_features/lightning_workflow/config/install/views.view.moderation_history.yml
+++ b/modules/lightning_features/lightning_workflow/config/install/views.view.moderation_history.yml
@@ -2,6 +2,7 @@ langcode: en
 status: true
 dependencies:
   module:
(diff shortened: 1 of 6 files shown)


modules/lightning_features/lightning_workflow/config/install/views.view.moderation_history.yml
```diff
@@ -339,7 +348,16 @@ display:
     position: 1
     display_options:
       display_extenders: {  }
-      path: node/%/moderation-history
+      path: node/%node/moderation-history
+      menu:
```
balsama: I don't think we had a link to this before. I agree it makes sense, I just wonder if it'll get unwanted attention if we add this now. Any particular motivation?

PR #542, opened 2017-12-13
Add package and datestamp to profile info file

balsama:
Given:

1. We don't think composer-built Lightning installs are getting properly added to the reported installs
2. There are two additional keys in the D.O-packaged info file

Let's just add those two keys.

### DIFF
--- a/src/Composer/ReleaseVersion.php
+++ b/src/Composer/ReleaseVersion.php
@@ -28,13 +28,13 @@ class ReleaseVersion {
     $finder = (new Finder())
       ->name('*.info.yml')
       ->in('.')
-      ->exclude('docroot');
+      ->exclude(['docroot', 'vendor']);
 
     /** @var \Symfony\Component\Finder\SplFileInfo $info_file */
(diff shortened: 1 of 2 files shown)


src/Composer/ReleaseVersion.php
```diff
@@ -34,6 +34,9 @@ public static function execute(Event $event) {
     foreach ($finder as $info_file) {
       $info = Yaml::parse($info_file->getContents());
 
+      if (@$info['name'] == 'Lightning' && $info['type'] == 'profile') {
```
phenaproxima: Why is $info['name'] error-suppressed?

balsama: Not all info files have the `name` and or `type` key. And rather than do two `isset`s, I figured it was acceptable to just suppress in this case.

balsama: ...turns out, this script is picking up the Console translations - those are the "things" without those keys. Will try to fix that instead.

PR #543, opened 2017-12-14
Add Console command that returns current version of Lightning

balsama:
We need this information in Lightning Project's CI for the "since" argument in our update-to-HEAD tests.

### DIFF
--- a/lightning.services.yml
+++ b/lightning.services.yml
@@ -45,7 +45,7 @@ services:
     tags:
       - { name: drupal.generator }
 
-  lightning.lightning_version:
+  lightning.version_command:
     class: 'Drupal\lightning\Command\VersionCommand'
     arguments:
             - '@app.root'
--- a/lightning.services.yml
+++ b/lightning.services.yml
@@ -44,3 +44,10 @@ services:
     arguments: []
     tags:
       - { name: drupal.generator }
+
+  lightning.lightning_version:
+    class: 'Drupal\lightning\Command\VersionCommand'
+    arguments:
+            - '@app.root'
+    tags:
+      - { name: drupal.command }
--- a/lightning.services.yml
+++ b/lightning.services.yml
@@ -48,6 +48,6 @@ services:
   lightning.version_command:
     class: 'Drupal\lightning\Command\VersionCommand'
     arguments:
-            - '@app.root'
+      - '@app.root'
     tags:
       - { name: drupal.command }
--- a/src/Command/VersionCommand.php
+++ b/src/Command/VersionCommand.php
@@ -8,12 +8,10 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Yaml\Yaml;
-use Drupal\Console\Annotations\DrupalCommand;
 
 /**
- * Class VersionCommand.
- *
- * @DrupalCommand
+ * A Drupal Console command that simply prints the semver converted version of
+ * the current Lightning codebase to stdout.
  */
 class VersionCommand extends Command {
 
@@ -38,13 +36,6 @@ class VersionCommand extends Command {
   /**
    * {@inheritdoc}
    */
-  protected function configure() {
-    $this->setName('lightning:version');
-  }
-
-  /**
-   * {@inheritdoc}
-   */
   protected function execute(InputInterface $input, OutputInterface $output) {
     $finder = (new Finder())
       ->files()
@@ -53,13 +44,16 @@ class VersionCommand extends Command {
 
     foreach ($finder as $info_file) {
       $info = Yaml::parse($info_file->getContents());
-      if (($info['distribution']['name'] == 'Lightning') && isset($info['version'])) {
+      if ($info['distribution']['name'] === 'Lightning' && isset($info['version'])) {
         // There should only be one file in docroot/profiles named
         // lightning.info.yml, but just to make sure we have the right file in
         // the iterator, break out when we have enough information to be
         // reasonably confident.
         break;
       }
+    }
+    if (!isset($info)) {
+      throw new \Exception('Lightning info file not found.');
     }
 
     $io = new DrupalStyle($input, $output);
@@ -76,7 +70,7 @@ class VersionCommand extends Command {
    *
    * NOTE: This will break if the minor version number is greater than 9.
    *
-   * @param $drupal_version
+   * @param string $drupal_version
    *   The version in 8.x-n.nn format.
    *
    * @return string
@@ -86,7 +80,10 @@ class VersionCommand extends Command {
     preg_match('/^8\.x-(\d+).(\d)(\d+)(-.+)?$/', $drupal_version, $matches);
     $semver = "$matches[1].$matches[2]." . intval($matches[3]);
     if (isset($matches[4])) {
-      $semver = $semver .$matches[4];
+      // $matches[4] is only populated if the version has a "-[prerelease]"
+      // string at the end so we must check to see if it exists before appending
+      // it back onto the end of the converted string.
+      $semver = $semver . $matches[4];
     }
     return $semver;
   }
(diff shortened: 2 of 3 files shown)


lightning.services.yml
```diff
@@ -44,3 +44,10 @@ services:
     arguments: []
     tags:
       - { name: drupal.generator }
+
+  lightning.lightning_version:
```
phenaproxima: This service ID is kinda ambiguous. Maybe lightning.version_command would be clearer?

src/Command/VersionCommand.php
```diff
@@ -0,0 +1,94 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Yaml\Yaml;
+use Drupal\Console\Annotations\DrupalCommand;
+
+/**
+ * Class VersionCommand.
+ *
+ * @DrupalCommand
```
phenaproxima: Do we need the DrupalCommand annotation?

src/Command/VersionCommand.php
```diff
@@ -0,0 +1,94 @@
...
+use Drupal\Console\Core\Command\Command;
+use Drupal\Console\Core\Style\DrupalStyle;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Yaml\Yaml;
+use Drupal\Console\Annotations\DrupalCommand;
+
+/**
+ * Class VersionCommand.
```
phenaproxima: Can this comment more clearly explain what the command does?

src/Command/VersionCommand.php
```diff
@@ -0,0 +1,94 @@
...
+  public function __construct($app_root) {
+    parent::__construct('lightning:version');
+    $this->appRoot = $app_root;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function configure() {
+    $this->setName('lightning:version');
```
phenaproxima: There's no need for this method; it's already covered in parent::__construct().

src/Command/VersionCommand.php
```diff
@@ -0,0 +1,94 @@
...
+   */
+  protected function execute(InputInterface $input, OutputInterface $output) {
+    $finder = (new Finder())
+      ->files()
+      ->name('lightning.info.yml')
+      ->in($this->appRoot . '/profiles');
+
+    foreach ($finder as $info_file) {
+      $info = Yaml::parse($info_file->getContents());
+      if (($info['distribution']['name'] == 'Lightning') && isset($info['version'])) {
```
phenaproxima: Nit: There's an extra set of parentheses here. Also, let's use === when comparing strings; sometimes PHP does weird shit.

src/Command/VersionCommand.php
```diff
@@ -0,0 +1,94 @@
...
+        // There should only be one file in docroot/profiles named
+        // lightning.info.yml, but just to make sure we have the right file in
+        // the iterator, break out when we have enough information to be
+        // reasonably confident.
+        break;
+      }
+    }
+
+    $io = new DrupalStyle($input, $output);
+    $io->info($this->toSemanticVersion($info['version']));
```
phenaproxima: There's no guarantee that $info actually exists. We should probably add an if (isset($info)) check here.

src/Command/VersionCommand.php
```diff
@@ -0,0 +1,94 @@
...
+   *   The version in 8.x-n.nn format.
+   *
+   * @return string
+   *   Semantic version
+   */
+  public static function toSemanticVersion($drupal_version) {
+    preg_match('/^8\.x-(\d+).(\d)(\d+)(-.+)?$/', $drupal_version, $matches);
+    $semver = "$matches[1].$matches[2]." . intval($matches[3]);
+    if (isset($matches[4])) {
+      $semver = $semver .$matches[4];
```
phenaproxima: Supernit: There should be an extra space after the period.

src/Command/VersionCommand.php
```diff
@@ -0,0 +1,94 @@
...
+   * @param $drupal_version
+   *   The version in 8.x-n.nn format.
+   *
+   * @return string
+   *   Semantic version
+   */
+  public static function toSemanticVersion($drupal_version) {
+    preg_match('/^8\.x-(\d+).(\d)(\d+)(-.+)?$/', $drupal_version, $matches);
+    $semver = "$matches[1].$matches[2]." . intval($matches[3]);
+    if (isset($matches[4])) {
```
phenaproxima, line 82: Can there be a small comment explaining what this does?

src/Command/VersionCommand.php
```diff
@@ -0,0 +1,94 @@
...
+   * Converts a Lightning release version to a semantic version number according
+   * to Lightning's VERSIONS.md file. Examples:
+   * - 8.x-1.23 => 1.2.3
+   * - 8.x-1.203 => 1.2.3
+   * - 8.x-1.230 => 1.2.30
+   * - 8.x-1.23-dev => 8.x-1.2.3-dev
+   *
+   * NOTE: This will break if the minor version number is greater than 9.
+   *
+   * @param $drupal_version
```
phenaproxima: Missing type hint.

src/Command/VersionCommand.php
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Converts a Lightning release version to a semantic version number according
+   * to Lightning's VERSIONS.md file. Examples:
+   * - 8.x-1.23 => 1.2.3
+   * - 8.x-1.203 => 1.2.3
+   * - 8.x-1.230 => 1.2.30
+   * - 8.x-1.23-dev => 8.x-1.2.3-dev
+   *
+   * NOTE: This will break if the minor version number is greater than 9.
```
phenaproxima, line 71: This is overzealous, but we could throw an InvalidArgumentException if, for some reason, the minor version number IS greater than 9 :)

tests/src/Kernel/VersionCommandTest.php
```diff
@@ -0,0 +1,32 @@
...
+    $drupal_versions = [
+      '8.x-1.23' => '1.2.3',
+      '8.x-1.203' => '1.2.3',
+      '8.x-1.230' => '1.2.30',
+      '8.x-1.23-dev' => '1.2.3-dev',
+    ];
+
+    foreach ($drupal_versions as $drupal_version => $expected_semantic_version) {
+      $generated_semantic_version = VersionCommand::toSemanticVersion($drupal_version);
+      $this->assertEquals($expected_semantic_version, $generated_semantic_version);
```
phenaproxima: Should be assertSame().

phenaproxima (changes requested):
Looks pretty good. Found nitpicks, as you might expect :)

src/Command/VersionCommand.php
balsama: I don't think it's possible to detect that - which is the crux of the problem. For example, given:

`8.x-1.111`, we have no way of knowing if that's supposed to be `1.1.11` or `1.11.1`, so we always assume it's `1.1.11`  - that is, we always assume the minor version is < 10.

The RegEx enforces this by mapping exactly one digit after the '`.`' to be the minor version.

lightning.services.yml
```diff
@@ -44,3 +44,10 @@ services:
     arguments: []
     tags:
       - { name: drupal.generator }
+
+  lightning.version_command:
+    class: 'Drupal\lightning\Command\VersionCommand'
+    arguments:
+            - '@app.root'
```
phenaproxima: Nit: Whitespace counts in YAML, so we need to outdent the argument a few times :)

phenaproxima (changes requested):
One small thing.

lightning.services.yml
balsama: darn good eye, sir.

PR #550, opened 2018-01-24
Update composer.json

grasmash:
Use caret to define minimum core version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -78,7 +78,7 @@
     },
     "require": {
         "cweagans/composer-patches": "^1.6.4",
-        "drupal/core": "~8.4.4",
+        "drupal/core": "^8.4.4",
         "drupal/lightning_api": "1.x-dev",
         "drupal/lightning_core": "1.x-dev",
         "drupal/lightning_layout": "1.x-dev",


phenaproxima:
Looks good to me.

However, I feel like we had the caret at some point, and switched it to the tilde for a reason. Therefore, I'd like @balsama to look this over before we merge it.

phenaproxima:
Related literature from @anavarre: https://www.drupal.org/node/2769841

balsama:
We use a tilde (and include the patch version) because Lightning 3.0.1 will (probably) not work core 8.5.x. Historically, we've always had to do at least some rework to make Lightning compatible with a new minor release of core. So 3.1.0 will be the first release that works with 8.5.x. [VERSIONS.md](https://github.com/acquia/lightning/blob/8.x-3.x/VERSIONS.md) documents this.

We do have a 8.5.x branch for testing ([2.4.x-dev](https://packagist.org/packages/acquia/lightning#2.4.x-dev)), but it's wildly out of date. So it would be best to just cut a new 3.1.x branch from HEAD of 8.x-3.x.

Lightning's components rely on ~8.4.n too though. So a 3.1.x branch of Lightning that required ~8.5.0 of core would always conflict with the components. I think these constraints should be loosened to use a caret.

### Proposed resolution
1. Loosen the core constraint on each component to ^8.4.0.
2. A new Lightning branch (3.1.x) that has ~8.5.0 constraint on core

I'll have more than 2 minutes in front of my computer tonight and can do this then.

phenaproxima:
I would suggest that we update each of the components first, to confirm that each individually works on 8.5.x. Then once all are updated, we can change Lightning's constraint.

balsama:
@grasmash See https://www.drupal.org/project/lightning/issues/2939509

PR #551, opened 2018-02-08
Use the new Lightning Dev infrastructure

phenaproxima:
Lightning Dev introduces a new system for configuring and running our Behat and PHPUnit tests. It took a lot of wrangling, but this PR changes the profile to use this infrastructure, and removes almost all tests that were moved to the components.


balsama:
There was a failure on travis that I restarted last night. Looks to have passed this time. Hopefully we're not introducing periodic failures now that the tests take 30+ minutes :/

Anyway, the pipelines job is failing because it still uses phing. The pipelines job should be simpler because it doesn't test upgrades, phpunit, or javascript tests. But we need it passing because it provides our nightly builds. Once the pipelines build passes, we should check the PR branch out on cloud and make sure the cloud hooks actually still install the site. I can help if needed with that.

hussainweb:
We were using some of these step definitions and upgrading to Lightning 3.0.2 breaks our tests. Are these step definitions available elsewhere? I checked acquia/lightning_dev but they were not present there.

We were using some simple ones like "I wait for _x_ seconds" and some little complex ones involving entity browsers.

sylus:
We also lost our CKEditor tests etc which were being leveraged.

zachomedia:
We were also using some of the steps, including "When I visit the edit form" and "Then CKEditor :arg1 should exist".

phenaproxima:
Moments ago, we moved these contexts into the 8.x-2.x branch of acquia/lightning-core. We're going to bring them into the 8.x-1.x branch as well, hopefully today (if tests pass and all that jazz). So this will be fixed too.

@zachomedia: Just FYI, "When I visit the edit form" is a special step that depends on a hook, and it *will not work* without Lightning Dev. To fix that, you'll want to add this hook to a custom module:

```php
/**
 * Implements hook_local_tasks_alter().
 */
function YOURMODULE_local_tasks_alter(array &$local_tasks) {
  $local_tasks['entity.node.edit_form']['options']['attributes']['rel'] = 'edit-form';
}
```

phenaproxima:
@hussainweb: Specialized contexts, like the ones relating to entity browsers, are NOT going to be in Lightning Core. They are part of Lightning Media, and will remain there. So if you have a dependency on Lightning Media, you should be able to autoload its subcontexts.

sylus:
Sorry to be a pain was curious if there was a possibility of a release today? Have a few sub profiles depending on us but can't release till newest Lightning alpha / stable tag is out.

Thanks for all the amazing work guys!

balsama:
@sylus New releases have already been made for the components (API, Core, and Media) so you should be able to grab those now since 3.0.2 doesn't pin the components.

balsama:
@sylus But yes, making the release now.

sylus:
Thanks everyone for the quick turnaround, your amazing :)

PR #554, opened 2018-02-16
Update Lightning Core

phenaproxima:
The failures at https://travis-ci.org/acquia/lightning/builds/341106804 are kinda inexplicable. I can't reproduce any of them locally. I propose we try updating Lightning Core, which underwent some changes in the last couple of days, and see if that smoothes things out.


phenaproxima:
Turns out that it's not necessary to do this, since CI always pulls the latest HEAD of the components. I'm closing this PR.

PR #556, opened 2018-03-02
Add instructions about how to use update:lightning to capture config

balsama:
In talking with @grasmash and @arknoll, it became clear that there was some confusion about what exactly the scope of `update:lightning` is and how it should be used. For example, `update:lightning` is described as running optional configuration updates. So it should technically only need to be run in you dev environment. But in 2.2.4, it triggered the migration to content moderation - which would also need to be run on production databases.

In that case, WBM2CM provided a Drush command that could do the same thing, but it's easy to see that the lines are being blurred.

Proposed resolution:
1. update:lightning` is to be used for configuration updates only (this was already the intention, but we should be strict about it in the future)
2. Migrations that can't happen in database updates will have to use another method. For example, the possible future migration to Layout Builder could be handled by a UI in the Panelizer module.

### DIFF
--- a/UPDATE.md
+++ b/UPDATE.md
@@ -21,6 +21,10 @@ For a typical site that has a properly configured directory for exporting config
 that is managed by your VCS, you would generally follow these steps:
 
 #### In your development or local environment.
+1. Read the [release notes](https://github.com/acquia/lightning/releases)
+   for the release to which you are updating, and any other releases between
+   your current version.
+   
 1. Update your codebase.
   
   ```
@@ -29,28 +33,28 @@ that is managed by your VCS, you would generally follow these steps:
   composer update acquia/lightning --with-all-dependencies
   ```
 
-2. Run any database updates.
+1. Run any database updates.
   
   ```
   drush cache-rebuild
   drush updatedb
   ```
   
-3. Run any Lightning configuration updates.
+1. Run any Lightning configuration updates.
   
   ```
   drush cache:rebuild
   drush update:lightning
   ```
   
-4. Export the new configuration.
+1. Export the new configuration.
 
   
   ```
   drush config-export
   ```
 
-5. Commit the code and configuration changes to your VCS and push them to your
+1. Commit the code and configuration changes to your VCS and push them to your
    destination environment.
   
 #### On your destination environment.
@@ -62,7 +66,7 @@ that is managed by your VCS, you would generally follow these steps:
   drush updatedb
   ```
 
-2. Import any configuration changes.
+1. Import any configuration changes.
   
   ```
   drush cache-rebuild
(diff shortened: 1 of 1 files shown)


UPDATE.md
```diff
@@ -16,6 +16,63 @@ the "Configuration updates" section of this file.
...
+#### In your development or local environment.
+1. Read the [release notes](https://github.com/acquia/lightning/releases)
+   for the release to which you are updating, and any other releases between
+   your current version.
+   
+1. Update your codebase.
+  
+  ```
+  composer self-update
+  composer require acquia/lightning:~3.1.0 --no-update
```
phenaproxima: How are we going to remember to change this version constraint in future minor releases?

phenaproxima (approved):
I like this. It's really nice and clear.

balsama:
@grasmash @arknoll Let me know if you have any feedback here. Trying to be more specific about how `update:lightning` should be used and what its scope is based on conversions with both of you.

UPDATE.md
```diff
@@ -16,6 +16,64 @@ the "Configuration updates" section of this file.
...
+  ```
+  composer self-update
+  composer require acquia/lightning:~[LIGHTNING_VERSION] --no-update
+  composer update acquia/lightning --with-all-dependencies
+  ```
+
+1. Run any database updates.
+  
+  ```
+  drush cache-rebuild
```
anavarre: Should be `cache:rebuild`

UPDATE.md
```diff
@@ -16,6 +16,64 @@ the "Configuration updates" section of this file.
...
+  ```
+  drush cache:rebuild
+  drush update:lightning
+  ```
+  
+1. Export the new configuration.
+
+  
+  ```
+  drush config-export
```
anavarre: Should be `config:export`

UPDATE.md
```diff
@@ -16,6 +16,64 @@ the "Configuration updates" section of this file.
...
+
+1. Commit the code and configuration changes to your VCS and push them to your
+   destination environment.
+  
+#### On your destination environment.
+
+1. Run any database updates.
+  
+  ```
+  drush cache-rebuild
```
anavarre: Should be `cache:rebuild`

UPDATE.md
```diff
@@ -16,6 +16,64 @@ the "Configuration updates" section of this file.
...
+  drush updatedb
+  ```
+
+1. Import any configuration changes.
+  
+  ```
+  drush cache-rebuild
```
anavarre: Should be `cache:rebuild`

UPDATE.md
```diff
@@ -16,6 +16,64 @@ the "Configuration updates" section of this file.
...
+  drush config-import
```
anavarre: Should be `config:import`

PR #559, opened 2018-03-07
3.1.0 release

balsama:


balsama:
Note that the fixture will need to be regenerated once core releases comes out.

PR #560, opened 2018-03-20
Avoid deprecated usage of user property

brockfanning:
This change avoids the deprecation error generated here: https://github.com/jhedstrom/drupalextension/blob/master/src/Drupal/DrupalExtension/Context/RawDrupalContext.php#L173

### DIFF
--- a/tests/features/bootstrap/entity.behat.inc
+++ b/tests/features/bootstrap/entity.behat.inc
@@ -74,7 +74,7 @@ class EntitySubContext extends DrupalSubContextBase {
     $node = $scope->getEntity();
 
     if (!isset($node->uid)) {
-      $user = $this->getContext(DrupalContext::class)->user;
+      $user = $this->getContext(DrupalContext::class)->getUserManager()->getCurrentUser();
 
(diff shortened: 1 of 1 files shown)


phenaproxima:
Looks good, but I think we can actually use $this->getUser() instead, which should be inherited from DrupalSubContextBase.

brockfanning:
@phenaproxima Thanks - just gave that a try as well, but that also gives me a deprecation error: `16384: DrupalSubContextBase::getUser() is deprecated. Use RawDrupalContext::getUserManager()->getCurrentUser() instead.`

phenaproxima:
Heh -- go figure! Thanks for the heads-up. I'm OK with merging this when tests pass. @balsama, do you concur?

brockfanning:
I don't suppose those test failures are a known issue?

phenaproxima:
I think those are random failures (they happen sometimes); restarting both failed builds.

brockfanning:
Looks like one of them failed again - might need another restart?

balsama:
Thanks @brockfanning!

PR #561, opened 2018-03-22
Remove unneeded methods in EntitySubContext

phenaproxima:
@anavarre ran into a problem where he got this error while using the Drush driver to interact with Drupal during Behat testing:

```
Notice: Undefined property: stdClass::$uid in /var/www/html/blt/docroot/profiles/contrib/lightning/tests/features/bootstrap/entity.behat.inc line 12
```

We don't test with the Drush driver and never have, so it's entirely possible that this is something normal that we've never run into. There are two solutions: we could either test with the Drush driver and investigate, or we could remove the offending method from EntitySubContext, because it is no longer necessary anyway. I chose the second path :)


balsama:
To keep the diff as small as possible, can we avoid the changes to composer.lock and *.make by not updating dependencies at all? Especially if someone might need to pull this in as a patch.

phenaproxima:
I don't think we can; without updating dependencies, Pipelines kept failing it because of the lock file being allegedly out of sync with composer.json.

phenaproxima:
I can easily create a patch that includes only the needed fixes, without any of the dependency stuff.

balsama:
OK. I added 00683ffc63066873524aae8bd790e55dc2ee9671 which updates Lightning API and JSON API and does the other cruft from this PR, so that they diff should be cleaner.

PR #566, opened 2018-03-26
Always build lock file with PHP 7

balsama:
We use our lock file to generate the Lightning Strict meta package. And the documentation there states that it work with php >= PHP 7. The problem is, if we generate Lightning's lock file with a local version of php 7.1 or 7.2, we end up with dependencies that aren't compatible with PHP 7.

Luckily, if we generate with PHP 7, the dependencies will also work with 7.1 and 7.2 (this might not always be the case, but it is now).

This PR will force dependencies to be resolved for PHP 7.0.8 regardless of the host's actual version of PHP. Why 7.0.8 specifically? Because drupal/core requires `php ^5.5.9|>=7.0.8`.

Note that this will prevent you from running the profile build from source on a machine with php 5.6. It _does not_ prevent you from running Lightning in the context of a scaffold project on php 5.6.


phenaproxima:
Looks good and right to me, but I wonder: should we maybe document somewhere that, as of this commit, you effectively need PHP 7.0.8 or later in order to work on Lightning itself?

balsama:
Good point. What if we added a "Contributing" section to the README?

README.md
```diff
@@ -134,10 +134,38 @@ for Drush which allow you to uninstall dependencies of parent profiles.
 
 * [Drush 9 inherited profile dependencies patch](https://www.drupal.org/files/issues/2902643-2--drush-master.patch).
 
+## Contributing
+Issues are tracked on [Drupal.Org][issue_queue]. Contributions can be provided
```
phenaproxima: Nit: Drupal.org should be drupal.org.

README.md
```diff
@@ -134,10 +134,38 @@ for Drush which allow you to uninstall dependencies of parent profiles.
...
+either as traditional patches or as pull requests on our [GitHub clone][github].
+
+Each Lightning component also has a Drupal.Org issue queue:
```
phenaproxima: Same here.

README.md
```diff
@@ -134,10 +134,38 @@ for Drush which allow you to uninstall dependencies of parent profiles.
...
+
+* [API](https://www.drupal.org/project/issues/lightning_api)
+* [Core](https://www.drupal.org/project/issues/lightning_core)
+* [Layout](https://www.drupal.org/project/issues/lightning_layout)
+* [Media](https://www.drupal.org/project/issues/lightning_media)
+* [Workflow](https://www.drupal.org/project/issues/lightning_workflow)
+
+### Developing Lightning Locally
```
phenaproxima: I'd change this to "Local Development".

README.md
```diff
@@ -134,10 +134,38 @@ for Drush which allow you to uninstall dependencies of parent profiles.
...
+Lightning has a dev dependency on Lightning Dev which provides tools to aid in
```
phenaproxima: Should we link to Lightning Dev?

README.md
```diff
@@ -134,10 +134,38 @@ for Drush which allow you to uninstall dependencies of parent profiles.
...
+devloping and contributing to Lightning. To install locally:
+
+1. Clone this repo
```
phenaproxima: Let's actually give the git clone and composer install commands here.

phenaproxima (changes requested):
Looks great -- just a few minor fixes.

phenaproxima (approved):

PR #567, opened 2018-03-26
Fix exposed filter name in documentation

balsama:
According to @malikkotob, the manual update instructions for Lightning Media are misleading. The filter name is actually "Media Type", the label is "Provider", and we want to change it to "Type".

It'd be good to confirm this by installing 3.0.3 before merging. And we should have a parallel PR against Lightning Media to change the annotation on Update200.php.

### DIFF
--- a/UPDATE.md
+++ b/UPDATE.md
@@ -167,8 +167,7 @@ follow the instructions for updating from 2.2.0 to 2.2.1, then from 2.2.1 to
 
 ### 3.0.3 to 3.1.0
 * Edit the **Media** view, and if it has an exposed filter called "Media Type",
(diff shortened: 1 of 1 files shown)


balsama:
I just installed 3.0.3. It looks like, indeed, the filter name is "Media Type":
media__media____drush_site-install

But the Label and Identifier are "Source" and "source", not "Provider/provider":
media__media____drush_site-install2

hmmm.

balsama:
I suspect that there were optional config changes to this view in that past. How about this... we can be purposefully ambiguous and not specify what the existing values are for Label and Identifier.

PR #569, opened 2018-03-27
Sub-profile utilities should link to documentation

phenaproxima:
It's an unambiguous UX and DX improvement. Need I say more?


balsama:
RTBC from me. We can merge this after 3.1.1 comes out on 28 March.

PR #572, opened 2018-04-02
Fix broken core branch in packaging script

phenaproxima:
I don't know how the hell this packaging script /ever/ worked. It was putting a GitHub URL into drupal-org-core.make, which breaks the packaging system on drupal.org. It was also misinterpreting the name of the core branch.

This should fix these problems and build correctly on drupal.org.


balsama:
I think this is failing because of the requirement on composer. See this issue for more info:
https://www.drupal.org/project/lightning/issues/2940386

phenaproxima:
Yes and no. It's failing because Composer brings in the JSON Schema validation library, which in turn is  leveraged by JSON API. And this breaks things because of JSON API and Metatag not playing nicely together.

The short term solution is to remove composer/composer from the dependencies, even though it is legitimate. We should fix this at some point, but it's more important to get the drupal.org build working correctly by fixing the packaging script.

PR #580, opened 2018-05-11
Use a shell script to generate tarballs

phenaproxima:
Currently, the process for generating working tarballs (i.e., for distribution on GitHub and possibly other venues) is cumbersome and requires a global copy of Drush 8. We can do a lot better than that.

This PR adds a shell script to generate the tarball:

```
$ ./tarball.sh 3.104
```
This will download a temporary copy of Drush 8, build Lightning, and archive it as lightning-8.x-3.104.tar.gz, which will install correctly on Cloud and other environments that are not Composer-ready. 
I tested this manually to confirm that the contents of the tarball do, in fact, install correctly.

Additionally, this PR guts build.xml and replaces every target we have with a deprecation message. We can remove all of that infrastructure in Lightning 4.x.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1,235 +1,60 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project name="Lightning" default="env">
 
-  <taskdef name="setxmlproperty" classpath="${project.basedir}/src/Task" classname="SetXMLPropertyTask" />
-
-  <!-- Locations of required binaries. -->
-  <property name="drush" value="${project.basedir}/vendor/bin/drush" />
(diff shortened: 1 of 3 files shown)


danepowell:
@phenaproxima if we've already deprecated everything in build.xml, can Lightning go ahead and remove it from the codebase? Or do we still need to wait for 4.x?

I ask because static code analysis tools for some reason tend to flag it as a vulnerability. It's clearly a false positive, but regardless, if it's unnecessary maybe it would be better to just dump it and stop fighting with these scanners.

PR #581, opened 2018-05-11
Fix extraneous Drupal core patch directories

danepowell:
This is the result of discussion in https://github.com/acquia/blt/pull/2809

This will have no effect today, but as soon as Composer Patches 1.6.5 or later is released, this configuration will stop the duplicate core/b and core/core directories appearing, as well as failing core patches.

We could update composer patches to 1.x-dev to take advantage of this fix today, but I assume that's not desirable.

I've tested to ensure this works with 1.x-dev.


phenaproxima:
I approve of this once tests are passing, but I'd like @balsama to sanity-check too.

balsama:
Nice! +1 from me. I'm glad there's movement in Composer Patches around this.

Pipelines doesn't rebuild the lock file, so it's going to complain that it doesn't match composer.json. As such, this PR should have a rebuilt lock file included too. But that's minor.

danepowell:
No problem, I rebuilt the lock file too. I wasn't sure how that worked with distros.

I just used `composer update --lock`, for some reason that removed the libraries key from the Drush makefile. I'm not sure if that was expected... ?

balsama:
Thanks Dane!
> removed the libraries key from the Drush makefile. I'm not sure if that was expected

Known issue with out packaging script unfortunately. We usually fix it on release commits.

balsama:
Tests are failing for other reasons. We'll get this committed shortly.

balsama:
Passing now. @phenaproxima This is RTBC from me.

amcgowanca:
Glad to see this coming through... I've been using a different approach to solve this for a handful of different projects. Thanks @danepowell !

phenaproxima:
Merged 15da466ba8f78f6b19c49c00149c8dbb7b4dff5b, restoring the libraries to drupal-org.make on commit. Thanks, @danepowell!

PR #583, opened 2018-06-01
Remove start Phantom from Pipelines instructions

balsama:
Lightning dev no longer provides Phantom. And we stopped testing JavaScript on Pipelines a long time ago. We should stop trying to start it before running our tests.

This PR also includes a quicker way to get a cloud-compatible version of Drush that doesn't require nearly as much Composer renegotiation.

### DIFF
--- a/acquia-pipelines.yml
+++ b/acquia-pipelines.yml
@@ -32,7 +32,7 @@ events:
             - drush runserver --default-server=builtin 8080 &>/dev/null &
             - sleep 10
             - cd $SOURCE_DIR
-            #- behat --stop-on-failure --config .behat.yml --tags=~javascript
+            - behat --stop-on-failure --config .behat.yml --tags=~javascript
       - cleanup:
           type: script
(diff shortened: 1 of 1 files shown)


acquia-pipelines.yml
```diff
@@ -30,15 +30,15 @@ events:
...
-            - behat --stop-on-failure --config .behat.yml --tags=~javascript
+            #- behat --stop-on-failure --config .behat.yml --tags=~javascript
       - cleanup:
           type: script
           script:
             - cd $SOURCE_DIR
             # Setup settings file and codebase with minimum required for cloud.
             - lightning configure:cloud
-            # Remove dev tools from the Cloud build.
-            - composer remove acquia/lightning_dev --dev
```
balsama, line 44: `composer remove` in theis scenario is rather expensive and was causing memory issues on Pipelines. The real goal is to not have Drush 9 in the build. So this accomplished that.

PR #588, opened 2018-08-02
Install Pathauto by default

csirmazbendeguz:
This PR enables the Pathauto module on install.

Corresponding Drupal.org issue:
https://www.drupal.org/project/lightning/issues/2976783


phenaproxima:
Looks good to me, but we need to fix that Pipelines failure. @balsama will follow up with some info.

balsama:
Looks like this is timing out on removing Drush in the cleanup step. I've gotten around this by just manually removing Drush like this:
https://github.com/acquia/lightning/blob/8.x-3.x/acquia-pipelines.yml#L43

The whole reason we're removing Drush is because I get get errors in our Cloud hooks which seem to get confused when they find the Drush9 executable in our codebase. It seems to me that Cloud is now compatible with Drush9 if you explicitly call the Drush9 executable instead of just `drush`. I did this with our website config yesterday. So maybe try calling Drush9 in the Cloud hooks first before resorting to the removing the executable. Example:
https://github.com/balsama/lightning-microsite/blob/3.0.x/hooks/common/post-code-update/post-code-update.sh#L10

phenaproxima:
@csirmazbendeguz, I don't know how the hell it could have happened, but can you confirm that, in your fork, acquia-pipelines.yml has the lines that forcibly remove Drush, as @balsama suggests? For whatever reason, Pipelines is timing out because it's trying to do it the "old way", where it uses Composer to remove lightning_dev.

csirmazbendeguz:
Yes, that was the issue. Merged the 8.x-3.x branch.

PR #590, opened 2018-08-07
Add optional config for Comment support

csirmazbendeguz:
This PR provides an optional default comment type.

**These configs are copied from the standard profile.**

In the standard profile these configs are not optional, so it also adds the comment field to the article content type by default - is this something we should do (requires custom code)?


csirmazbendeguz:
This is failing for unknown reasons. It passed locally. Maybe a rerun is necessary?

phenaproxima:
The failure is due to Pathauto :) We'll have to fix Lightning Layout so that it tests Pathauto integration too. I'll deal with that; until I've got that fixed, this is blocked.

phenaproxima:
Fixed the Pathauto integration in Lightning Layout, so we should be able to rerun these now.

phenaproxima:
Okay, so -- to get this working, I need you to make a small change to lightning.profile. In `lightning_modules_installed()`, you'll see this:

```php
    entity_create('node_type', [
      'type' => 'article',
      'name' => 'Article',
    ])->save();
```

This should be removed; Lightning Workflow now creates that content type for testing when Lightning Dev is enabled.

csirmazbendeguz:
Done. This fixes the assertion that fails: https://github.com/acquia/lightning-media/pull/45

phenaproxima:
So one question I have here, is why did the media assertion fail for only two of five builds? It should have failed for all of them, I'd think. I'm restarting the broken builds; if this is a random failure, I think we need to squash it before we commit.

tests/src/Functional/DefaultCommentTypeTest.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Drupal\Tests\lightning\Functional;
+
+use Drupal\comment\Entity\CommentType;
+use Drupal\Tests\BrowserTestBase;
+
+/**
+ * @group lightning
+ */
+class DefaultCommentTypeTest extends BrowserTestBase {
```
phenaproxima: Thank you for writing this test, but I think we can actually remove it. I say that for two reasons: first, it installs Lightning, which will make it incredibly slow; and second, it's really just testing the optional config system in core, which is already very well-covered by core's tests. So there's no need for it here. :)

phenaproxima (changes requested):

csirmazbendeguz:
It seems like pathauto - or the media patterns - is not getting enabled.

phenaproxima (approved):

phenaproxima:
Looks good to me. Let's not bother adding comments to any specific node types right now; that could be a can of worms we don't want to get into. But at least we now have the basics of Comment support in place. Thanks, Bendi!

PR #591, opened 2018-08-07
3.2.0-beta1 Release

phenaproxima:


UPDATE.md
```diff
@@ -170,6 +170,33 @@ are currently running 2.2.0 and are trying to update to 2.2.6, you will need to
 follow the instructions for updating from 2.2.0 to 2.2.1, then from 2.2.1 to
 2.2.2, in that order.
 
+### 3.1.5 to 3.2.0-beta1
+* If you have any sub-profiles (regardless of whether or not they extend
```
balsama, line 174: We have a script that does this, right? Shouldn't we mention it here?

phenaproxima: No need; it's a standard update:lightning script. If they're already using that, it'll be done for them. The manual update steps are just for people who, for whatever reason, don't want to do it automatically.

phenaproxima:
Let it be noted that I need to disable Pathauto temporarily for this release, because the other components do NOT test with Pathauto, even though they support it. Although the integration does work, the tests are broken because they didn't account for Pathauto. I will fix the broken components and re-enable Pathauto in the back to dev stage.

PR #592, opened 2018-08-08
Rename "save to media library" field

csirmazbendeguz:
Renames labels in test config files. Prerequisite: renaming labels in lightning_media first.

### DIFF
--- a/tests/config/field.field.media.document.field_media_in_library.yml
+++ b/tests/config/field.field.media.document.field_media_in_library.yml
@@ -9,7 +9,7 @@ id: media.document.field_media_in_library
 field_name: field_media_in_library
 entity_type: media
 bundle: document
-label: 'Save to my media library'
+label: 'Show in media library'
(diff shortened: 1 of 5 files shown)


phenaproxima (changes requested):
This looks good to me, but it will need an update class. See Drupal\lightning\Update\Update320 for an example. The update, if executed, should make these same changes to every instance of field_media_in_library.

That's in addition to changing the code which actually creates new instances of this field, as well as the instances of it that we ship with default config.

csirmazbendeguz:
Ok, I can't find the Drupal\lightning\Update\Update320 class, but I found other Update classes.

csirmazbendeguz:
I added the Update class to the lightning_media PR. I think it belongs there, because this one only changes the tests' configs.

phenaproxima:
This was fixed in Lightning Media today, closing as duplicate.

PR #595, opened 2018-08-27
Config installer fix

csirmazbendeguz:
https://www.drupal.org/project/lightning/issues/2994679


balsama:
Thanks! Yeah, the hook_install limitation is a somewhat sloppy workaround to the fact that people misuse it. The Shortcuts implementation in core's Standard is a perfect example. So core decided to just outright ban it.

Thanks for this. My first thought is that we should probably clean up the stuff that was in there - especially the frontpage view config - rather than just dump it wholesale into a different file. But I'll let @phenaproxima  take a look.

phenaproxima:
So I'm not sure I'm on board with adding Yet Another Module to Lightning, just to do install tasks. My suggestion would be to implement hook_install_tasks(), which should (according to its documentation) allow us to run callback functions to do all this work during the actual install process. Can you look into that as a possible solution?

phenaproxima:
I think my suggestion will work: https://travis-ci.org/phenaproxima/lightning/builds/421154695. It installs cleanly with config_installer (and under local testing with Drush), and all our config changes are actually made on a normal, clean install (verified by tests).

phenaproxima:
I'm taking care of this in #596.

PR #604, opened 2018-09-08
Move install tasks to hook_install_tasks()

balsama:


phenaproxima:
This looks great. I think we only need to fix line 163 of lightning.profile and then this will pass (and be mergeable).

PR #610, opened 2018-10-09
3.2.1 Release

phenaproxima:


phenaproxima:
CI is failing due to a PHPUnit testing change which is present in 8.6.x HEAD of Drupal core, but not in a stable release (yet). The relevant change record (https://www.drupal.org/node/2998888) says this will be fixed in 8.6.2.

So, I'm merging this, because it effectively passed.

PR #611, opened 2018-10-10
Prevent the drupal.org tarball from being installed

phenaproxima:
Finally. It's time for us to cripple the d.o tarball, because it has been non-functional for a long time anyway.

This PR adds a new hidden module as a dependency, called lightning_install, whose only purpose is to check Lightning's info file for the string "# Information added by Drupal.org packaging script". If that string is there, hook_requirements() fails, explaining that Lightning cannot be installed from the tarball, and directing users to the Lightning Project repository. (This had to be its own module because Drupal's installer does not run hook_requirements() for profiles.)


balsama:
As part of this PR, can we provide the exact wording for the D.O project page.

balsama:
Manual tests:
* Run exact script that D.O runs and confirm resulting tarball is what we expect (that is, there should be a warning on the installer)
* Run the cloud tarball script and confirm that a tarball which contains vendor in docroot (and all of Lightning) is created - and is installable.

Let's upload the two resulting tarballs here so others can test too.

phenaproxima:
Proposed wording for the d.o page:

The tarball below **does not install** because drupal.org does not package Lightning's Composer dependencies. Visit [our GitHub repository](https://github.com/acquia/lightning/releases) to download a working tarball, or [install Lightning with Composer](https://github.com/acquia/lightning-project).

balsama:
> Visit our GitHub repository to download a working tarball
Does it make sense to provide and link to a Lightning Project Composer-generated tarball at this point? Only Cloud needs the makefile generated one. If we must provide a tarball (and I think we must), why not make it something people could actually sustainably use?

Preference for creating new projects:
1. `composer create-project acquia/lightning project`
2. Download composer-generated tarball (functionally equivilent to \#1)
3. Download a makefile-generated tarball (perhaps we don't even need to link this one?)

phenaproxima:
Let's just reverse the order of actions here. First, link to Lightning Project (which has no tarball, and never has). Then we link to the Cloud-ready tarballs for people who MUST use a tarball for some reason. And that's it (we have no other type of tarball that will work).

So:

The tarball below **does not install** because drupal.org does not package Lightning's Composer dependencies. [Install Lightning with Composer](https://github.com/acquia/lightning-project) instead. If you *must* use a tarball, you can download one from [our GitHub repository](https://github.com/acquia/lightning/releases).

phenaproxima:
Confirmed in manual testing that everything works as expected. Cloud tarballs still install properly, and d.o generated tarballs will be built out as normal, but the lightning_install module will throw up the requirements wall.

phenaproxima:
Here is a Cloud-ready tarball to test with (GitHub will not accept files larger than 10 MB here): (deleted)

phenaproxima:
And here is a drupal.org-equivalent tarball to test with: (deleted)

balsama:
Great. Confirmed that the hook_requirements error also prevents installation from Drush. We can merge this.

Please update the D.O project page with the text from the comment above. Let's also add links to the issues that block us from using D.O's packaging script in that text.

> The tarball below does not install because drupal.org does not package Lightning's Composer dependencies (See #2763017 and #2715441). Install Lightning with Composer instead. If you must use a tarball, you can download one from our GitHub repository.

phenaproxima:
Merged into 8.x-3.x and cherry-picked to 8.x-4.x.

weitzman:
Hurray!

PR #612, opened 2018-10-15
Add support for Drupal profiles living outside of d.o.

miromichalicka:
We want to use Lightning packaging mechanisms for packaging our Lightning-based profile. The profile is hosted on GitHub and it's registered on regular Packagist. This PR brings changes, which won't ignore packages which are outside of `drupal/` composer namespace.

### DIFF
--- a/src/Composer/Package.php
+++ b/src/Composer/Package.php
@@ -153,6 +153,7 @@ class Package {
       case 'drupal-core':
       case 'drupal-theme':
       case 'drupal-module':
+      case 'drupal-profile':
         $info['type'] = substr($package['type'], 7);
         break;
     }
@@ -212,6 +213,8 @@ class Package {
         '$1.$2$3',
         $version
       );
+    }
+    if (strpos($package['name'], 'drupal/') === 0) {
       unset($info['download']);
     }
     return $info;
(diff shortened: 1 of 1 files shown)


phenaproxima:
Thanks for this pull request!

I'll be very honest: although I'd like @balsama's input here, I'm quite hesitant to merge this. The reason is that the Package.php script in Lightning is emphatically _not_ meant for outside use, and eventually we are hoping to remove it. It is only there to help us package Lightning for distribution on drupal.org -- a mechanism which has been deprecated for a while and is now borderline unsupported. We're trying to go Composer-only once drupal.org allows it (we're pretty much already there, TBH), so any further changes to the packaging script is not something I'd like to encourage. My suggestion would be to just copy the script wholesale into your profile and use/modify it at your leisure, which will give you what you need without requiring us to indefinitely maintain this piece of infrastructure.

That said, if there is enough pushback, and @balsama approves, I can probably be convinced to merge it. But I thought I'd put my 2 cents in now and get the discussion rolling.

miromichalicka:
@phenaproxima thanks for feedback. We aren't happy about need for this script either. We need it for Acquia Cloud package unfortunately.

PR #613, opened 2018-10-16
Don't run profiles tests

csirmazbendeguz:
Patch [#1356276](https://www.drupal.org/project/drupal/issues/1356276?page=1) brings in two tests tagged with the "profiles" group.
They will fail, because Lightning is a distribution, and when a distribution is present it will automatically get selected instead of the testing profile.
The solution is to not run these tests.


balsama:
I believe this was resolved by removing the distribution key. Can we close?

PR #614, opened 2018-10-17
3.2.2 Release

phenaproxima:


phenaproxima:
The failures are due to a change in Drupal core's testing infrastructure which will be in 8.6.3 (or the next bugfix release), but is not in the current security release. Behat tests passed on all builds, so this is an acceptable failure. Merging.

PR #615, opened 2018-10-18
3.1.8 Release

phenaproxima:


phenaproxima:
The failures are, as with #614, due to upstream changes in core's testing infrastructure. I'm not sure it will be fixed in the 8.5.x branch of core, since I don't know if any further non-security releases will be cut from that.

In any event, all Behat tests passed. This is an acceptable failure, so I'm merging it.

PR #622, opened 2018-11-15
Remove hook_install() from generated subprofiles

danepowell:
Lightning doesn't have a hook_install() implementation in the base profile, I think specifically so that it will be compatible with the new D8.6 feature of installing site from existing config.

However, the subprofile template still creates a hook_install() implementation, which breaks the ability to install sites from existing config.

An example of this breaking things: https://github.com/acquia/blt/pull/3241#issuecomment-439105307

### DIFF
--- a/templates/profile/install.twig
+++ b/templates/profile/install.twig
@@ -4,11 +4,3 @@
  * @file
  * Install, update and uninstall hooks for the {{ profile }} subprofile.
  */
-
-/**
- * Implements hook_install().
- */
-function {{ machine_name }}_install() {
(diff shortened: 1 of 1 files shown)


phenaproxima:
+1 for removing the function, but to compensate for that, let's keep the comment explaining that hook_install() should be implemented in order to do various tasks. But let's also mention that it will render the profile incompatible with config installs, and cite a change record to back that up.

danepowell:
I added the requested comments. I don't think the test failures are specific to this PR.

balsama:
Rerunning tests. RTBC from me if they pass - which they should.

PR #625, opened 2018-12-18
Temporarily ignore tests that are broken due to refactoring in components

csirmazbendeguz:


phenaproxima:
Can you also exclude 368f0045? `behat --tags '~@e787aec5' --tags '~@368f0045'` oughta do it.

PR #628, opened 2019-01-14
Behat needs to use Lightning Workflow's FixtureContext to pass

csirmazbendeguz:


phenaproxima:
The PHPUnit runs fail, but that's not related to this pull request (it has to do with Acquia Connector and Facets, sigh). But the Behat failure is dealt with quite nicely. I'm merging this.

PR #640, opened 2019-04-13
Adding simple telemetry module

grasmash:
This PR:

1. Adds a new lightning_telemetry module, which is disabled by default.
1. Adds an opt-in form to the Lightning profile installer. This enables the module if the user opts in.
1. Creates a cron hook that will send anonymized version data to Amplitude. Failures are silent by design.

Event data looks like this:
```
    "drupal.version": "8.7.0-dev",
    "extensions.lightning.status": "enabled",
    "extensions.lightning.version": "8.x-4.x-dev",
    "extensions.lightning_api.status": "enabled",
    "extensions.lightning_api.version": "8.x",
    "extensions.lightning_core.status": "enabled",
    "extensions.lightning_core.version": "8.x",
    "extensions.lightning_dev.status": "disabled",
    "extensions.lightning_dev.version": "8.x-2.300-dev",
    "extensions.lightning_layout.status": "disabled",
    "extensions.lightning_layout.version": "8.x",
    "extensions.lightning_media.status": "disabled",
    "extensions.lightning_media.version": "8.x",
    "extensions.lightning_workflow.status": "enabled",
    "extensions.lightning_workflow.version": "8.x",
    "php.version": "7.2.8"
```


grasmash:
@phenaproxima would you mind reviewing this? If it passes review, do _not_ merge yet. Need to finalized the opt-in language first.

grasmash:
Travis CI failures appear to be unrelated to this pull request.

balsama:
Thanks for doing this! At first glance, this should be in Lightning Core, not the profile.

lightning.profile
```diff
@@ -23,6 +23,11 @@ function lightning_install_tasks() {
   $tasks['lightning_set_logo'] = [];
   $tasks['lightning_alter_frontpage_view'] = [];
 
+  $tasks['Drupal\lightning\Form\TelemetryOptInForm'] = [
```
phenaproxima: Nit: can we use the `::class` syntax here?

lightning.profile
```diff
@@ -23,6 +23,11 @@ function lightning_install_tasks() {
...
+    'display_name' => t('Telemetry Opt-in'),
```
phenaproxima: We might wanna bikeshed the wording of this a bit.

modules/lightning_telemetry/CONTRIBUTING.md
```diff
@@ -0,0 +1,4 @@
+Failures to send Lightning Telemetry are intended to be graceful and quiet. Exceptions are caught and output is intentionally suppressed.
+
+To disable error suppression, set the environmental variable LIGHTNING_TELEMETRY_LOUD to a value that evaluates as true.
+E.g., `LIGHTNING_TELEMETRY_LOUD=1 drush cron`.
```
phenaproxima: Ideally this would be a configuration value, not an environment variable.

modules/lightning_telemetry/README.md
```diff
@@ -0,0 +1,4 @@
+# Lightning Telemetry
+
+This modules sends anonymized data about Drupal and Lightning usage
```
phenaproxima: Should be "This module sends anonymous data..."

modules/lightning_telemetry/lightning_telemetry.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'Lightning Telemetry'
+core: 8.x
+type: module
+package: Lightning
+description: 'Provides Acquia with anonymized telemetry data for Lightning usage.'
```
phenaproxima: s/anonymized/anonymous. Also, should we say "usage" instead of "telemetry"?

modules/lightning_telemetry/lightning_telemetry.module
```diff
@@ -0,0 +1,36 @@
...
+function lightning_telemetry_cron() {
+  /** @var \Drupal\lightning_telemetry\Telemetry $telemetry_service */
+  $telemetry_service = \Drupal::service('lightning_telemetry.telemetry');
+  $telemetry_service->sendTelemetry('Drupal cron ran');
+}
+
+/**
+ * Implements hook_modules_installed().
+ */
+function lightning_telemetry_modules_installed($modules) {
```
phenaproxima: $modules should be type hinted as an array.

modules/lightning_telemetry/lightning_telemetry.module
```diff
@@ -0,0 +1,36 @@
...
+  /** @var \Drupal\lightning_telemetry\Telemetry $telemetry_service */
+  $telemetry_service = \Drupal::service('lightning_telemetry.telemetry');
+  $installed_lightning_extensions = array_intersect($modules, $telemetry_service->getLightningExtensionNames());
```
phenaproxima: I think Lightning Core includes a class, called ComponentDiscovery, which is responsible for locating all Lightning components in a system. We should use that, rather than hard-coding a list into the telemetry service.

modules/lightning_telemetry/lightning_telemetry.module
```diff
@@ -0,0 +1,36 @@
...
+  if ($installed_lightning_extensions) {
+    $event_properties = ['installed_extensions' => array_values($installed_lightning_extensions)];
+    $telemetry_service->sendTelemetry('Lightning extensions installed', $event_properties);
+  }
+}
+
+/**
+ * Implements hook_modules_uninstalled().
+ */
+function lightning_telemetry_modules_uninstalled($modules) {
```
phenaproxima: $modules should be type hinted as an array.

balsama:
@phenaproxima Can you still review given my above comment?

modules/lightning_telemetry/lightning.services.yml
```diff
@@ -0,0 +1,4 @@
+services:
```
phenaproxima: This file should not exist. Since it's in the lightning_telemetry module, it will never be read. The services file should be lightning_telemetry.services.yml.

modules/lightning_telemetry/lightning_telemetry.services.yml
```diff
@@ -0,0 +1,4 @@
+services:
+  lightning_telemetry.telemetry:
```
phenaproxima: Since this is part of the Lightning suite, I think we can call this service `lightning.telemetry` to save ourselves some typing.

modules/lightning_telemetry/lightning_telemetry.services.yml
```diff
@@ -0,0 +1,4 @@
...
+    class: Drupal\lightning_telemetry\Telemetry
+    arguments: ['@extension.list.module', '@http_client', '@config.factory']
```
phenaproxima: Because we are using `extension.list.module`, this module will NOT work on Drupal 8.5.x or earlier. We should document that, _and_ we should add a hook_requirements() implementation to prevent installation on those old versions.

modules/lightning_telemetry/src/Event.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+namespace Drupal\lightning_telemetry;
+
+/**
+ * AmplitudeEvent.
```
phenaproxima: This doc comment seems kinda pointless?

modules/lightning_telemetry/src/Event.php
```diff
@@ -0,0 +1,81 @@
...
+    return $this;
+  }
+
+  /**
+   * Converts the event object to a JSON string that Amplitude will accept.
+   *
+   * @return string
+   *   JSON string that Amplitude will accept.
+   */
+  public function __toJson() {
```
phenaproxima: This is implementing a magic method -- does that come from some interface, or...?

modules/lightning_telemetry/src/Event.php
```diff
@@ -0,0 +1,81 @@
...
+    return json_encode(
```
phenaproxima: Let's use Drupal\Component\Serialization\Json::encode() here.

modules/lightning_telemetry/src/Event.php
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * Sets $this->data.
+   *
+   * @param array $data
+   *   The event properties.
+   *
+   * @return $this
+   *   The event object.
+   */
+  public function setData($data) {
```
phenaproxima: $data should be type hinted as an array.

modules/lightning_telemetry/src/Event.php
```diff
@@ -0,0 +1,81 @@
...
+   * An array of event data.
+   *
+   * This key value pair will be sent to Amplitude as event_properties.
+   *
+   * @var array
+   */
+  protected $data;
+
+  /**
+   * The Amplitude user id for the event.
```
phenaproxima: What is Amplitude? I think we might need more documentation around this. Also, "id" should be "ID" everywhere.

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Telemetry service.
+ */
+class Telemetry {
+
+  /**
+   * Amplitude API URL.
+   */
+  const AMPLITUDE_API_URL = 'https://api.amplitude.com/httpapi';
```
phenaproxima: These constants should probably be configuration values.

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,188 @@
...
+   * side code. Fetching data requires an additional API secret.
+   */
+  const AMPLITUDE_API_KEY = 'f32aacddde42ad34f5a3078a621f37a9';
+
+  /**
+   * The module handler to invoke the alter hook with.
+   *
+   * @var \Drupal\Core\Extension\ModuleHandlerInterface
+   */
+  protected $moduleHandler;
```
phenaproxima: We can remove this property, as it is not used anywhere.

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,188 @@
...
+   * The HTTP client.
+   *
+   * @var \GuzzleHttp\ClientInterface
+   */
+  protected $httpClient;
+
+  /**
+   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   */
+  protected $config_factory;
```
phenaproxima: Needs a (useless) description in the comment. :)

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  /**
+   * Sends an event to Amplitude.
+   *
+   * @param \Drupal\lightning_telemetry\Event $event
+   *   The Amplitude event.
+   *
+   * @return bool
+   *   TRUE if the request to Amplitude was successful.
```
phenaproxima: "...successful, FALSE otherwise."

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,188 @@
...
+   */
+  protected function sendEvent(Event $event) {
+    $response = $this->httpClient->request('POST', self::AMPLITUDE_API_URL, [
+      'form_params' => [
+        'api_key' => self::AMPLITUDE_API_KEY,
+        'event' => $event->__toJson(),
```
phenaproxima: We should generally avoid calling magic methods directly. Presumably, using \Drupal\Component\Serialization\Json::encode() will automatically call the __toJson() method.

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,188 @@
...
+      ]
+    ]);
+    $success = $response->getStatusCode() == 200;
```
phenaproxima: Nit: Let's just `return $response->getStatusCode() === 200`.

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,188 @@
...
+   * @param string $type
+   *   The event type.
+   * @param array $properties
+   *   The event properties.
+   *
+   * @return \Drupal\lightning_telemetry\Event
+   *   An Amplitude event with basic info already populated.
+   */
+  protected function createEvent($type, $properties) {
+    $user_id = $this->config_factory->get('system.site')->get('uuid');
```
phenaproxima: Okay, so...we said that the data is anonymous. Although this is not _personally_ identifiable, isn't this maybe a bit more intrusive than we should be? During module installation, we could just generate a _new_ UUID, stored in our own configuration, to anonymize this even further. Not sure if that would be overkill, though.

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,188 @@
...
+    $default_properties['extensions'] = $this->getExtensionInfo();
+    $default_properties['php']['version'] = phpversion();
+    $default_properties['drupal']['version'] = \Drupal::VERSION;
+    $properties = array_merge_recursive($default_properties, $properties);
```
phenaproxima: I think we should probably use Drupal\Component\Utility\NestedArray::mergeDeep() here.

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,188 @@
...
+    return $event;
+  }
+
+  /**
+   * Gets an array of all Lightning Drupal extensions.
+   *
+   * @return array
+   *   A flat array of all Lightning Drupal extensions.
+   */
+  public function getLightningExtensionNames(): array {
```
phenaproxima: This should use the ComponentDiscovery class from Lightning Core, rather than hard-code a list of extensions which we _will_ forget to update.

src/Form/TelemetryOptInForm.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @package Drupal\lightning\Form
+ */
+class TelemetryOptInForm extends FormBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'telemetry_opt_in';
```
phenaproxima: This should be prefixed by `lightning_`.

src/Form/TelemetryOptInForm.php
```diff
@@ -0,0 +1,65 @@
...
+   */
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    if ($form_state->getValue('allow_telemetry')) {
+      \Drupal::service('module_installer')->install(['lightning_telemetry']);
```
phenaproxima: The module installer should be injected. (Forms have a static `create()` method for this purpose.)

phenaproxima (dismissed):
Initial review looks pretty good; I have a number of concerns, but nothing serious. I think this looks sane overall.

phenaproxima:
Agreed with @balsama that this module should be a subcomponent of Lightning Core. The opt-in form should be part of the Lightning profile, but that's the _only_ part of this PR which needs to be in the profile. So maybe we can iterate here, get the module looking right, and then open a new PR against Lightning Core 3.x. We can also cherry-pick it into Lightning Core 4.x.

One thing we are missing here is tests. All we need is a kernel test or two to prove that the telemetry events are sent at the right times, and that they contain the data we expect; we can accomplish that with some reasonably simple service mocking.

grasmash:
@phenaproxima Kernel tests and Service mocking are not my strong point, would you be able to point out an example? I likely have little time to directly develop now that sprints are over, but basic refactoring should be ok.

modules/lightning_telemetry/lightning.services.yml
grasmash: I got a CI error when I left this file out.

modules/lightning_telemetry/lightning_telemetry.services.yml
grasmash: Can't we just add this to lightning_telemetry.info.yml?

```
dependencies:
  - drupal:system (>= 8.6)
```

modules/lightning_telemetry/src/Event.php
grasmash: It's just required for coding standards.

modules/lightning_telemetry/src/Event.php
grasmash: It's not a magic method. Would you prefer it not resemble one?

modules/lightning_telemetry/src/Telemetry.php
grasmash: They should never change though. Why not make them constants?

modules/lightning_telemetry/src/Telemetry.php
grasmash: This is not a magic method, and it doesn't look like \Drupal\Component\Serialization\Json::encode() will call it.

modules/lightning_telemetry/src/Telemetry.php
grasmash: Maybe we should just change the language to say PII? It feels like overkill to make another UUID which is no more private than the other. I mean, what's the difference?

modules/lightning_telemetry/src/Telemetry.php
grasmash: Will that return exactly this array?

modules/lightning_telemetry/CONTRIBUTING.md
grasmash: Eh. Like a real CM one? I'm really only looking at this for debugging. I don't think a customer would ever use it.

modules/lightning_telemetry/src/Telemetry.php
balsama: > what's the difference?

One could be correlated to the other if we use the UUID. It's possible that the UUID could leak with identifying information elsewhere which would correlate this data to the identifying info. Seems like a good move to store a new UUID or hash this one before sending.

balsama:
@TravisCarden Did a research spike on this last quarter. It'd be good to get his thoughts on this approach.

My first thought is that this should live in Acquia Connector with implementation in Lightning (and other Drupal Product modules) if needed. Acquia Connector is installed on far more sites than Lightning the profile is.

I'm also still questioning the value of adding something like this to a module which needs to be enabled. @lcatlett mentioned that much (all?) of this information might be available using some existing tools. @lcatlett can you weigh in?

grasmash:
@balsama I'd be happy to move this into connector, but I'd prefer to first get a lightweight and working implementation in Lightning first, then refactor into a more general purpose use case later. Frankly, this is a proof of value.

modules/lightning_telemetry/src/Telemetry.php
grasmash: If we move this into Acquia Connector or into a standalone module in the future, we may may want to extend this array to include more Acquia product modules. So, perhaps it's best not to optimize this for the near-term use case.

modules/lightning_telemetry/src/Telemetry.php
grasmash: Well, a hash sounds simplest so let's go with that.

modules/lightning_telemetry/CONTRIBUTING.md
TravisCarden: I would suggest the State system. Environment variables aren't durable or convenient enough, and configuration appears in configuration exports and travels between environments, which you don't want.

lightning.profile
TravisCarden: Do we need to run the language by Legal?

I propose something like the "Allow Lightning to send anonymous usage data to Acquia".

balsama: I believe @grasmash is doing this:

> do not merge yet. Need to finalized the opt-in language first.

So that string might change.

modules/lightning_telemetry/CONTRIBUTING.md
TravisCarden: Also, why is this in `CONTRIBUTING.md`? It seems more like README material.

src/Form/TelemetryOptInForm.php
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $form['#title'] = "Telemetry Opt-in";
+    $form['allow_telemetry'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Allow Lightning to send anonymous telemetry data to Acquia'),
+      // @todo Revise and finalize language.
+      '#description' => t('This module sends anonymous data about Acquia product usage to Acquia for product development purposes. No private information will be gathered. Data will not be used marketing
```
TravisCarden: Missing a word:
```diff
-      '#description' => t('This module sends anonymous data about Acquia product usage to Acquia for product development purposes. No private information will be gathered. Data will not be used marketing
+      '#description' => t('This module sends anonymous data about Acquia product usage to Acquia for product development purposes. No private information will be gathered. Data will not be used for marketing
```

src/Form/TelemetryOptInForm.php
```diff
@@ -0,0 +1,94 @@
...
+    return new static(
+      $container->get('module_installer')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $form['#title'] = "Telemetry Opt-in";
```
TravisCarden: UI standards stipulate sentence case for headings:
```diff
-    $form['#title'] = "Telemetry Opt-in";
+    $form['#title'] = "Telemetry opt-in";
```

lightning.profile
TravisCarden: In case the string happens to stay the same, UI standards stipulate sentence case for headings:
```diff
-    $form['#title'] = "Telemetry Opt-in";
+    $form['#title'] = "Telemetry opt-in";
```

modules/lightning_telemetry/README.md
```diff
@@ -0,0 +1,15 @@
+# Lightning Telemetry
+
+This module sends anonymous data about Acquia product usage to Acquia
+for product development purposes.
+
+No private information will be gathered. Data will not be used marketing
```
TravisCarden: Missing a word:
```diff
-No private information will be gathered. Data will not be used marketing
+No private information will be gathered. Data will not be used for marketing
```

modules/lightning_telemetry/README.md
```diff
@@ -0,0 +1,15 @@
...
+and will not be sold to any third parties.
+
+Telemetry is opt-in only and can be disabled at any time by uninstalling
+the lightning_telemetry module.
+
+
+# Requirements
+
+This module requires Drupal 8.5.x or later.
```
TravisCarden: Per this PR, `lightning_install.info.yml` is requiring 8.6.x or later. I hesitate to duplicate details like version numbers in documentation that easily gets out of sync. Do we really need a "Requirements" section in a README at all?

modules/lightning_telemetry/lightning_telemetry.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'Lightning Telemetry'
+core: 8.x
+type: module
+package: Lightning
+description: 'Provides Acquia with anonymous telemetry data for Lightning usage.'
```
TravisCarden: We should use the same description everywhere:
```diff
- description: 'Provides Acquia with anonymous telemetry data for Lightning usage.'
+ description: 'Sends anonymized data about Drupal and Lightning usage for product development purposes.'
```

modules/lightning_telemetry/lightning_telemetry.module
```diff
@@ -0,0 +1,36 @@
+<?php
+
```
TravisCarden:
```diff
+ /**
+  * @file
+  * Install, update, and uninstall functions for the Lightning Telemetry module.
+  */
+ 
```

modules/lightning_telemetry/CONTRIBUTING.md
grasmash: Because it's really only for people developing this module. I would never expect an end user to use this variable.

grasmash:
Since we may move this module to the acquia_connector, I'd like to actually rename it to acquia_telemetry. That would simplify the upgrade path greatly.

modules/lightning_telemetry/lightning_telemetry.services.yml
TravisCarden: That seems like it should work fine. We should _also_ add a requirement or conflict to `lightning_telemetry/composer.json`.

modules/lightning_telemetry/src/Telemetry.php
TravisCarden: I agree that unless there are different endpoints (like dev > stage > prod, or something) there's no reason to make this configurable.

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,194 @@
...
+        'api_key' => self::AMPLITUDE_API_KEY,
+        'event' => Json::encode($event),
+      ]
+    ]);
+
+    return $response->getStatusCode() == 200;
+  }
+
+  /**
+   * Creates and sends a cron event to Amplitude.
```
TravisCarden:
```diff
-   * Creates and sends a cron event to Amplitude.
+   * Creates and sends an event to Amplitude.
```

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,194 @@
...
+   *
+   * @param string $event_type
+   *   The event type.
```
TravisCarden: What are the acceptable event type values?

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,194 @@
...
+   *
+   * @param array $event_properties
+   *   Event properties.
+   *
+   * @throws \Exception
+   */
+  public function sendTelemetry($event_type, array $event_properties = []) {
```
TravisCarden: If we allow arbitrary properties to be sent, anybody could use this method to populate our metrics with garbage. We probably need to consider a properties whitelist approach.

TravisCarden: ...garbage _values_ to whitelisted properties could be a problem, too. Depending on our level of concern, we might need to be even more aggressive, such as type testing values or somehow preventing non-Acquia code from sending metrics at all.

modules/lightning_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,194 @@
...
+    $default_properties['extensions'] = $this->getExtensionInfo();
+    $default_properties['php']['version'] = phpversion();
+    $default_properties['drupal']['version'] = \Drupal::VERSION;
+    $properties = NestedArray::mergeDeep($default_properties, $properties);
+    $event = new Event($type, $user_id, $properties);
+
+    return $event;
+  }
+
+  protected function getUserId() {
```
TravisCarden: Missing docblock.

src/Form/TelemetryOptInForm.php
```diff
@@ -0,0 +1,94 @@
...
+
+use Drupal\Core\Extension\ModuleInstallerInterface;
+use Drupal\Core\Form\FormBase;
+use Drupal\Core\Form\FormStateInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Class TelemetryOptInForm.
+ *
+ * This form is displayed during the Lightning install process. It provides
```
TravisCarden, line 16: I don't see this form being surfaced anywhere _after_ initial installation, so we're not likely to get any pre-existing users to opt in. Should we add some kind of `hook_requirements()` implementation that prompts the user to install the module or hide the message?

modules/acquia_telemetry/CONTRIBUTING.md
```diff
@@ -0,0 +1,4 @@
+Failures to send telemetry are intended to be graceful and quiet. Exceptions are caught and output is intentionally suppressed.
```
TravisCarden:
```diff
+ # Contribution Guide
+
 Failures to send telemetry are intended to be graceful and quiet. Exceptions are caught and output is intentionally suppressed.
```
...or something.

modules/acquia_telemetry/README.md
```diff
@@ -0,0 +1,10 @@
+# Acquia Telemetry
+
+This module sends anonymous data about Acquia product usage to Acquia
+for product development purposes.
+
+No private information will be gathered. Data will not be used for
+marketing and will not be sold to any third parties.
+
+Telemetry is opt-in only and can be disabled at any time by uninstalling
```
TravisCarden, line 9: The only concern I have about having the `acquia_telemetry` installation _constitute_ opting into telemetry is that this is a consciously short-term solution. Will replacing it with our permanent solution be more complicated by this approach? Should we use a State value instead?

TravisCarden (commented):
My feedback is in.

lcatlett (commented):
@phenaproxima I was able to confirm that we can already generate a lot of this data with existing tools and I made some updates, but those reports are dependent on AHT commands and running them as a user or bot that has both AHT and CCI API access. The biggest pain point that I have seen with module based approaches that require the user to opt-in to the module is that we tend to discourage enabling these type of modules in production due to possible performance implications. There are other approaches that don't require an additional module or bootstrapping Drupal, but I'm not really clear on what data this module is intending to collect that we can't access already via the existing support and hosting methods, which would then have the added benefit of inheriting the bastion/aht access controls to that data. @grasmash are you aware of any GDPR, FedRamp, PCI compliance and/or HIPAA implications for previous amplitude integrations?

grasmash:
> are you aware of any GDPR, FedRamp, PCI compliance and/or HIPAA implications for previous amplitude integrations?

Yes I am aware. This has been reviewed from those standpoints, and GDPR compliance is exactly why this is opt-in.

> those reports are dependent on AHT commands and running them as a user or bot that has both AHT and CCI API access. 

Exactly. We're focusing on measuring what we can where we can. A platform level solution would be better, and we can pursue that. But, there is still value is using this as a stop-gap.

modules/lightning_telemetry/src/Telemetry.php
grasmash: I'm not very concerned with receiving garbage values. Amplitude provides a way to ignore these when/if it happens. I'm not sure this warrants the effort on this side of things.

src/Form/TelemetryOptInForm.php
grasmash: It's a good idea. Maybe we should, but I wouldn't consider it MVP.

modules/acquia_telemetry/README.md
grasmash: To me, tying Telemetry to installation seems to simplify things. It provides a simple interface for enabling/disabling the behavior, and we can be sure that no bugs cause accidental capturing of data.

grasmash:
@phenaproxima @TravisCarden @lcatlett I've addressed all feedback with either a code change or a comment. Does this pass review?

modules/lightning_telemetry/src/Telemetry.php
grasmash: I've changed this to getAcquiaExtensionNames() to make this more clear.

modules/lightning_telemetry/lightning_telemetry.services.yml
TravisCarden: @grasmash we still need a `lightning_telemetry/composer.json` _with a requirement or conflict_ per ^^.

modules/lightning_telemetry/CONTRIBUTING.md
phenaproxima: I don't know if we need a module-specific CONTRIBUTING.md. I'm +1 for @TravisCarden's idea of using a state setting to disable error suppression.

modules/lightning_telemetry/src/Telemetry.php
phenaproxima: I still think we need to use ComponentDiscovery to find the extensions. We are not going to remember to update this obscure array if we add new subcomponents.

lightning.profile
```diff
@@ -23,6 +23,11 @@ function lightning_install_tasks() {
   $tasks['lightning_set_logo'] = [];
   $tasks['lightning_alter_frontpage_view'] = [];
 
+  $tasks[Drupal\lightning\Form\TelemetryOptInForm::class] = [
```
phenaproxima: By `::class` syntax, I meant we should have a `use` statement at the top of the file, then this array key can be a much more readable `TelemetryOptInForm::class`. :)

modules/acquia_telemetry/README.md
phenaproxima: I agree with @grasmash on this one. Assuming it is disabled by default, installing the module is a very clear opt-in action, and we don't need to further complicate it.

modules/acquia_telemetry/src/Event.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+namespace Drupal\acquia_telemetry;
+
+use JsonSerializable;
```
phenaproxima: `JsonSerializable` is in the global namespace, so there is no need for this `use` statement. The `implements` clause will need to be changed to `\JsonSerializable`, though.

modules/acquia_telemetry/src/Event.php
```diff
@@ -0,0 +1,88 @@
...
+  protected $type;
+
+  /**
+   * An array of event data.
+   *
+   * This key value pair will be sent to Amplitude as event_properties.
+   *
+   * @var array
+   */
+  protected $data;
```
phenaproxima: This should be an empty array by default.

modules/acquia_telemetry/src/Event.php
```diff
@@ -0,0 +1,88 @@
...
+namespace Drupal\acquia_telemetry;
+
+use JsonSerializable;
+
+/**
+ * An event object as defined by Amplitude API specifications.
+ *
+ * @see https://developers.amplitude.com/#http-api
+ */
+class Event implements JsonSerializable {
```
phenaproxima: Let's make this class `final`, and all the internal (currently protected) things `private`. We want to draw strong boundaries around this class. Core doesn't allow private right now, but because we are Lightning we can happily ignore that policy. :)

modules/acquia_telemetry/src/Event.php
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Event constructor.
+   *
+   * @param string $type
+   *   The event type.
+   * @param array $data
+   *   The event properties.
+   */
+  public function __construct($type, $user_id, $data = []) {
```
phenaproxima: The doc comment does not mention $user_id, and $data should be type hinted as an array.

modules/acquia_telemetry/src/Event.php
```diff
@@ -0,0 +1,88 @@
...
+   *   An array, to be serialized json  that the Amplitude API will accept.
+   *
+   * @see https://amplitude.zendesk.com/hc/en-us/articles/204771828#keys-for-the-event-argument
+   */
+  public function jsonSerialize() {
+     return [
+        'event_type' => $this->type,
+        'user_id' => $this->user_id,
+        'event_properties' => $this->data,
+      ];
```
phenaproxima: I think the closing square bracket is indented a space too far...

modules/acquia_telemetry/src/Event.php
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * Converts the event object to a serializable array for the Amplitude API.
+   *
+   * This method will be be called by PHP whenever json_encode() is used on
+   * an object of this class.
+   *
+   * @return array
+   *   An array, to be serialized json  that the Amplitude API will accept.
```
phenaproxima: 'json' should be "JSON", since it's an acronym.

modules/acquia_telemetry/src/Event.php
```diff
@@ -0,0 +1,88 @@
...
+   *   The event object.
+   */
+  public function setData(array $data) {
+    $this->data = $data;
+
+    return $this;
+  }
+
+  /**
+   * Converts the event object to a serializable array for the Amplitude API.
```
phenaproxima: I think the entire doc comment can simply say "Implements \JsonSerializable::jsonSerialize()."

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+use Drupal\Component\Utility\Crypt;
+use Drupal\Component\Utility\NestedArray;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Extension\ModuleExtensionList;
+use GuzzleHttp\ClientInterface;
+
+/**
+ * Telemetry service.
+ */
+class Telemetry {
```
phenaproxima, line 17: Let's make this class final, and all its protected members private. This is another class we should draw sharp boundaries around. (In general, that's the policy I'm taking towards Lightning development; classes should be final and private unless they have a specific reason not to be -- this allows us to have _intentional_ APIs, rather than accidental ones.)

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * Amplitude API URL.
+   *
+   * @see https://developers.amplitude.com/#http-api
+   */
+  const AMPLITUDE_API_URL = 'https://api.amplitude.com/httpapi';
```
phenaproxima: I'm not a huge fan of these constants for two reasons:
1) They're exposed to outside code. It's not possible to make them private until PHP 7.1, which we don't yet support, sadly.
2) They're basically impossible to override for testing. This is the reason I wanted them as simple config values; if they were, we could unit test this class, and simulate HTTP requests, by passing in different config values. That said, this is a lesser concern because we _can_ mock the HTTP client and simulate the HTTP requests that way.

I'm not sure we _really_ need to change these, seeing as how we're not exactly designing a long-term API; just calling it out. One compromise, to fix the visibility issue, is for these to be `private static` properties.

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+   */
+  protected function sendEvent(Event $event) {
+    $response = $this->httpClient->request('POST', self::AMPLITUDE_API_URL, [
+      'form_params' => [
+        'api_key' => self::AMPLITUDE_API_KEY,
+        'event' => Json::encode($event),
+      ]
+    ]);
+
+    return $response->getStatusCode() == 200;
```
phenaproxima: In general, we should prefer ===.

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+   * @param string $event_type
+   *   The event type. This accepts any string that is not reserved. Reserved
+   *   event types include: "[Amplitude] Start Session", "[Amplitude] End
+   *   Session", "[Amplitude] Revenue", "[Amplitude] Revenue (Verified)",
+   *   "[Amplitude] Revenue (Unverified)", and "[Amplitude] Merged User".
+   *
+   * @param array $event_properties
+   *   Event properties.
+   *
+   * @throws \Exception
```
phenaproxima, line 127: Needs an explanation of why the exception would be thrown.

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+    ]);
+
+    return $response->getStatusCode() == 200;
+  }
+
+  /**
+   * Creates and sends an event to Amplitude.
+   *
+   * @param string $event_type
+   *   The event type. This accepts any string that is not reserved. Reserved
```
phenaproxima, line 117: Perhaps the Event constructor should validate that the event type is not a reserved type, and throw \InvalidArgumentException if it is?

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+   *   An array of extension info keyed by the extensions machine name.
+   */
+  protected function getExtensionInfo() {
+    $all_modules = $this->extensionListModule->getAllAvailableInfo();
+    $acquia_extensions = array_intersect_key($all_modules, array_flip($this->getAcquiaExtensionNames()));
+    $extension_info = [];
+
+    foreach ($acquia_extensions as $name => $extension) {
+      // Version is unset for dev versions.
+      $version = $extension['version'] ? $extension['version'] : $extension['core'];
```
phenaproxima: Needs to be `$version = isset($extension['version']) ? ... : ...`

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+   *
+   * @param \Drupal\Core\Extension\ModuleExtensionList $extension_list_module
+   *   The extension.list.module service.
+   * @param \GuzzleHttp\ClientInterface $http_client
+   *   The HTTP client.
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The config.factory service.
+   */
+  public function __construct(ModuleExtensionList $extension_list_module, ClientInterface $http_client, ConfigFactoryInterface $config_factory) {
+    $this->extensionListModule = $extension_list_module;
```
phenaproxima: Nit: I think $extensionListModule is an awkward name for the property. How about just $moduleList?

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+
+    foreach ($acquia_extensions as $name => $extension) {
+      // Version is unset for dev versions.
+      $version = $extension['version'] ? $extension['version'] : $extension['core'];
+      $extension_info[$name]['version'] = $version;
+    }
+
+    $installed_modules = $this->extensionListModule->getAllInstalledInfo();
+    foreach ($acquia_extensions as $name => $extension) {
+      if (array_key_exists($name, $installed_modules)) {
```
phenaproxima: Nit: This if/else bit could be a ternary.

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+    return $event;
+  }
+
+  /**
+   * Gets a unique ID for this application. "User ID" is an Amplitude term.
+   *
+   * @return string
+   */
+  protected function getUserId() {
+    return Crypt::hashBase64($this->config_factory->get('system.site')->get('uuid'));
```
phenaproxima: _Nice._

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  /**
+   * Gets an array of all Acquia Drupal extensions.
+   *
+   * @return array
+   *   A flat array of all Acquia Drupal extensions.
+   */
+  public function getAcquiaExtensionNames(): array {
```
phenaproxima: I still think we should use ComponentDiscovery to do this. If we add a subcomponent (or a top-level one), we _will_ forget to update this obscure array.

modules/acquia_telemetry/src/Telemetry.php
```diff
@@ -0,0 +1,205 @@
...
+    $acquia_extension_names = [
+      'lightning',
+      'lightning_api',
+      'lightning_core',
+      'lightning_dev',
```
phenaproxima: Lightning Dev is a defunct internal build/testing tool and should not be here.

modules/lightning_install/lightning_install.info.yml
```diff
@@ -5,3 +5,5 @@ hidden: true
 package: Lightning
 description: 'Provides helpful under-the-hood functionality used while installing Lightning.'
 version: '8.x-4.000-beta1'
+dependencies:
+  - drupal:system (>= 8.6)
```
phenaproxima: Why was this added to lightning_install.info.yml? This essentially forces all of Lightning to be 8.6+.

src/Form/TelemetryOptInForm.php
phenaproxima: One option is to add a hook_update_N() which simply sets a message asking users to enable the new module. That'd be very easy, and unobtrusive.

src/Form/TelemetryOptInForm.php
```diff
@@ -0,0 +1,94 @@
...
+ * the opportunity for users to opt-in to enabling the Lightning Telemetry
+ * module.
+ *
+ * @package Drupal\lightning\Form
```
phenaproxima: Lightning doesn't use `@package`.

src/Form/TelemetryOptInForm.php
```diff
@@ -0,0 +1,94 @@
...
+   */
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    if ($form_state->getValue('allow_telemetry')) {
+     $this->moduleInstaller->install(['lightning_telemetry']);
```
phenaproxima: The module is now acquia_telemetry. :) The fact that this almost went uncaught means that it needs a test. (A kernel test would work.)

phenaproxima (dismissed):

modules/lightning_telemetry/CONTRIBUTING.md
grasmash: As a developer, how would I quickly set the state from the CLI?

grasmash: Ah, `drush sset`.

modules/acquia_telemetry/src/Telemetry.php
grasmash: The Amplitude API should throw its own informative error if a reserved string is used. I highly doubt someone will use one of these event types, but I can add some validation if you think it's valuable.

modules/acquia_telemetry/src/Telemetry.php
grasmash: Trouble with making this final is that you can't then create a prophecy for the call when testing. Any suggestions?

balsama:
I opened this for discussion and posterity: https://backlog.acquia.com/browse/LG-142 (For when we ask ourselves what this is doing in Lightning in a years time 😉)

PR #641, opened 2019-04-18
Updating legal language.

grasmash:


src/Form/TelemetryOptInForm.php
```diff
@@ -23,6 +24,13 @@ final class TelemetryOptInForm extends FormBase {
    */
   private $moduleInstaller;
 
+  /**
+   * The extension.list.module service.
+   *
+   * @var \Drupal\Core\Extension\ModuleExtensionList
+   */
+  protected $moduleExtensionList;
```
phenaproxima: Can this be changed to private?

modules/acquia_telemetry/acquia_telemetry.info.yml
```diff
@@ -2,6 +2,14 @@ name: 'Acquia Telemetry'
 core: 8.x
 type: module
 package: Lightning
-description: 'Sends anonymized data about Drupal and Acquia product usage for product development purposes.'
+description: "By installing the Acquia Telemetry module, you agree to allow it to send anonymized telemetry data to Acquia for the purposes and on the conditions set forth below. If you do not agree, do not install this software.
+
+The Acquia Telemetry module sends anonymized data about Acquia product usage to Acquia for security, operations management, statistical analyses, and for research and development purposes. Specifically, it sends the following data:<br />
```
phenaproxima: Should we maybe include some P tags around the non-list parts of this, to make it a consistent block of HTML? Additionally, could you post a screenshot of how this renders?

phenaproxima (changes requested):
Looks good, would like a couple of small changes though...

phenaproxima (approved):

PR #650, opened 2019-05-16
Update, expand, simplify, and optimize automated testing setup (3.x)

TravisCarden:


.travis.yml
```diff
@@ -1,66 +1,58 @@
...
+    - { name: "Isolated test w/ recommended package versions", env: ORCA_JOB=ISOLATED_RECOMMENDED }
+    - { name: "Integrated test w/ recommended package versions", env: ORCA_JOB=INTEGRATED_RECOMMENDED }
+    - { name: "Integrated test w/ recommended package versions & previous minor release of Drupal core", env: ORCA_JOB=CORE_PREVIOUS }
+    - { name: "Isolated test w/ dev package versions", env: ORCA_JOB=ISOLATED_DEV }
+    - { name: "Integrated test w/ dev package versions", env: ORCA_JOB=INTEGRATED_DEV }
+    - { name: "Integrated test w/ dev package versions & next minor dev version of Drupal core", env: ORCA_JOB=CORE_NEXT }
+    # Custom job.
+    - { name: "Upgrade test", env: ORCA_JOB=CUSTOM ORCA_CUSTOM_FIXTURE_INIT_ARGS="--dev --no-site-install" ORCA_CUSTOM_TESTS_RUN_ARGS="--sut-only" DB_FIXTURE=3.0.0 }
+  allow_failures:
+    - env: ORCA_JOB=ISOLATED_DEV
```
phenaproxima: I don't think ISOLATED_DEV should be allowed to fail. :)

composer.json
```diff
@@ -4,6 +4,7 @@
     "description": "The best of Drupal, curated by Acquia",
     "license": "GPL-2.0-or-later",
     "require": {
+        "ext-dom": "*",
```
phenaproxima, line 7: Why did we add this?

src/Command/SubProfileCommand.php
```diff
@@ -404,9 +404,7 @@ protected function excludeSubComponents(array $parent, DrupalStyle $io) {
    * {@inheritdoc}
    */
   protected function execute(InputInterface $input, OutputInterface $output) {
-    $io = new DrupalStyle($input, $output);
-
-    if ($this->confirmGeneration($io)) {
+    if ($this->confirmOperation()) {
```
phenaproxima, line 407: Why was this changed?

phenaproxima (changes requested):
Couple of questions and one change. Feel free to prefix the next commit message with `[skip ci]` so that Travis CI won't run. We already know this passes. :)

composer.json
TravisCarden: Because I noticed there's code that uses the DOM extension. (PhpStorm helpfully highlights any code that uses a PHP extension that's not required in `composer.json`.)

src/Command/SubProfileCommand.php
TravisCarden: Because `confirmGeneration()` is deprecated and thus failing the deprecation scan. Happily, all the deprecated function does is call the new one, and it doesn't take the `$io` argument.

.travis.yml
TravisCarden: Removed.

phenaproxima (approved):

PR #655, opened 2019-07-10
Exclude test fixtures and other dev cruft from releases

danepowell:
This follows the process [described here](https://madewithlove.be/gitattributes/) to exclude dev code (especially, huge test fixtures) from stable Lightning releases on Github, which should also exclude them from Packagist.

Our use case is to try to improve project creation times for ADS users (and by extension, any BLT or Lightning user).

We determined that if you were to zip up an entire Drupal / Lightning codebase into a compressed tarball, it would be about 100MB in size. 50MB of that would consist _solely of Lightning's test fixtures_ (including Lightning Workflow, Lightning API, etc...). So there's quite an opportunity for optimization here.

If this process works for the Lightning meta package, I'd love to see it extended to the other Lightning packages.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+/git-hooks export-ignore
+/hooks export-ignore
+/tests export-ignore
+/.travis.yml export-ignore
+/acquia-pipelines.yml export-ignore
+/tarball.sh export-ignore


danepowell:
To verify that this works, see a tagged release I created via Github on this feature branch: https://github.com/danepowell/lightning/releases/tag/exclude-fixtures

phenaproxima:
Yup. I love it. I'm going to merge this when Travis CI OKs it.

TravisCarden:
I love this idea! In fact, I have it on my own list of things to look into. But at this moment, doing it will prevent ORCA from running Lightning tests in other products' builds. There's a way to change that, but it involves a small change to ORCA itself. Let me see if I can whip something up on that side. Please don't merge this yet.

TravisCarden:
Okay, I've modified ORCA to always install Acquia packages from source (https://github.com/acquia/orca/pull/31). That way ORCA builds will still get and be able to run tests even after they've been excluded from package releases.

You may merge this now. We may want to coordinate releases somewhat to minimize the impact on product builds.

I'm going to recommend this pattern (i.e., excluding non-production files from releases) as a standard for all Acquia packages. Is there a d.o issue to do this to Drupal core itself?

danepowell:
I think this is the core issue you are looking for, to reduce the core project size: https://www.drupal.org/project/drupal/issues/2338671

danepowell:
I posted an infra issue to see whether this can be replicated for Lightning (and Acquia's) packages on d.o: https://www.drupal.org/project/infrastructure/issues/3067445

I suspect the same .gitattributes trick might work but can't find any documentation on this.

PR #658, opened 2019-09-06
Add Autosave Form and Conflict modules

kdruckman:


src/Update/Update405.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+
```
phenaproxima: There's an extra blank line here that shouldn't be. :)

src/Update/Update405.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+final class Update405 implements ContainerInjectionInterface {
+
+  /**
+   * The module installer.
+   *
+   * @var \Drupal\Core\Extension\ModuleInstallerInterface
+   */
+  private $moduleInstaller;
+
```
phenaproxima, line 22: Same here.

src/Update/Update405.php
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Enables the Autosave Form and Conflict modules.
+   *
+   * @update
+   *
+   * @ask Do you want to enable the Autosave Form and Conflict modules?
+   */
+  public function enableAutosaveForm() {
+    $this->moduleInstaller->install(['autosave_form', 'conflict']);
+  }
```
phenaproxima, line 51: Inverse problem: there should be an extra blank line after the last closing brace of the last method. :)

phenaproxima (changes requested):
Looks fantastic, just needs tests and (literally) teensy whitespace changes. +++

phenaproxima (approved):

tests/src/Kernel/Update405Test.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     */
+  public function testEnableAutosaveForm() {
+
+    $moduleHandler = $this->container->get('module_handler');
+
+    $this->assertFalse($moduleHandler->moduleExists('autosave_form'));
+    $this->assertFalse($moduleHandler->moduleExists('conflict'));
+
+    $updateHandler = Update405::create($this->container);
```
phenaproxima: This line, and the next, can be combined into one line: `Update405::create($this->container)->enableAutosaveForm()`

tests/src/Kernel/Update405Test.php
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @group lightning
+ */
+class Update405Test extends KernelTestBase
+{
+    protected static $modules = array('system');
+
+    /**
+     */
+  public function testEnableAutosaveForm() {
```
phenaproxima, line 19: The doc comment of this method should probably be `@covers ::enableAutosaveForm`

tests/src/Kernel/Update405Test.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Drupal\Tests\lightning\Kernel;
+
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\lightning\Update\Update405;
+
+/**
+ * @group lightning
+ */
+class Update405Test extends KernelTestBase
```
phenaproxima, line 12: I think we should add a `@coversDefaultClass \Drupal\lightning\Update\Update405` to the doc comment of this class.

phenaproxima (changes requested):
I would only request very minor changes. I'll fix the coding standards and formatting on commit. :) Otherwise this looks great, nice work! 👍

phenaproxima (approved):
Looks good! I'm going to give this a quick manual test, then it's merge o'clock.

phenaproxima:
OK, so! Now that Lightning Workflow 3.9 is out, to bring this functionality in we will need to update composer.json so that our requirement for `drupal/lightning_workflow` is `^3.9` (i.e., "at least 3.9"). Will you make that change, @kdruckman? Also, when you do that, remember to run `composer update`, and commit automatic changes to `composer.lock` and any of the *.make files. :)

phenaproxima:
I manually tested this and it worked great both on a clean installation, and after updating from the previous version. So I cleaned up a few coding standards things, updated the documentation, and committed this in 04340f7936d2bd8dedfec797ecd5f60e25f26b33. Beautiful!

PR #659, opened 2019-09-23
Add the Redirect module to Lightning

kdruckman:
add redirect module


phenaproxima:
I think you might need to get in sync with the upstream remote, since you're still working from a fork. Do these instructions help? https://help.github.com/en/articles/syncing-a-fork

If not, let's handle this after scrum. :)

tests/src/Kernel/Update405Test.php
```diff
@@ -21,16 +21,34 @@ class Update405Test extends KernelTestBase {
...
+    $moduleHandler = $this->container->get('module_handler');
+
+    $this->assertTrue($moduleHandler->moduleExists('autosave_form'));
+    $this->assertTrue($moduleHandler->moduleExists('conflict'));
+  }
+
+  /**
+   * @covers ::enableRedirect
+   */
+  public function testEnableRedirect() {
```
phenaproxima: So...this actually doesn't need to be in its own method. I would suggest we rename testEnableAutosaveForm() to testUpdate(), and have it do something like this:

```php
$moduleHandler = $this->container->get(...);

$this->assertFalse($moduleHandler->moduleExists('autosave_form'));
// Same thing for Conflict and Redirect

$update = Update405::create($this->container);
$update->enableAutosaveForm();
$update->enableRedirect();

$moduleHandler = ...
$this->assertTrue($moduleHandler->moduleExists('autosave_form'));
// ...etc.
```
This allows us to test everything we need to, and keep things a bit simpler and less verbose. :)

phenaproxima (changes requested):
This will need a CHANGELOG entry, and it's blocked on a new release of Lightning Core (which will require an other go of `composer update`). Otherwise, it looks excellent overall; I would like a small optimization in the test. :)

tests/src/Kernel/Update405Test.php
kdruckman: makes sense!

tests/src/Kernel/Update405Test.php
```diff
@@ -19,18 +19,25 @@ class Update405Test extends KernelTestBase {
...
   /**
    * @covers ::enableAutosaveForm
+   * @covers ::enableRedirect
    */
-  public function testEnableAutosaveForm() {
-    /** @var \Drupal\Core\Extension\ModuleHandlerInterface $module_handler */
-    $module_handler = $this->container->get('module_handler');
-    $this->assertFalse($module_handler->moduleExists('autosave_form'));
-    $this->assertFalse($module_handler->moduleExists('conflict'));
+  public function testUpdate() {
```
phenaproxima, line 24: I think that there might be an extra blank line here that shouldn't be.

phenaproxima (approved):
I literally only found one _very_ minor thing wrong with this. It looks perfect otherwise.

tests/src/Kernel/Update405Test.php
kdruckman: I'll get those coding standards things figured out asap. :)

phenaproxima:
This worked great under quick manual testing. The CI failures are due to Chrome being a pain in the ass, and are not related. I ran the test locally and confirmed it passes. This is done, and I'm merging it. First-class work, @kdruckman, thanks!

PR #661, opened 2019-11-19
Add a machine-readable file containing information about previous versions

phenaproxima:
This addresses PF-2011.

It adds a releases.yml file at the top of the Lightning repository. This file will *not* be exported by git, which means it can only be accessed via the repo. It contains basic information for the RAs and other members of Acquia's support organization to help update customers' Lightning installations.

Each version is keyed by semantic version, and has a maximum of three keys:

`drupal_version`: The version number of this release, as it appears on drupal.org and in the administrative backend.
`has_manual_updates`: A boolean value indicating if this release needed manual updates. In most cases, these updates are also available as scripts that can be run by `drush update:lightning`.
`update_instructions`: If `has_manual_updates` is `true`, this links to a URL that explains how to do the update(s) by hand. This may also be present for releases that do not have manual update steps as such, but do have quirks (or anything else pertinent or peculiar) which may be of interest to RAs.


shaysmith (approved):
Looks good!

codexmas (approved):
Looks great. I can see RA-UP will need a little bit of version parsing code to handle the shifts from older versions to the newer ones and the 'beta1' type variants.

balsama (approved):

PR #662, opened 2019-12-06
Move infrastructure packages into dev dependencies

phenaproxima:


TravisCarden (approved):
This looks right to me.

composer.json
```diff
@@ -5,21 +5,21 @@
     "license": "GPL-2.0-or-later",
     "require": {
         "ext-dom": "*",
-        "cweagans/composer-patches": "^1.6.4",
```
danepowell, line 8: `cweagans/composer-patches` is required to apply patches defined by other Lightning projects. I agree that it doesn't make much sense as a required package here since Lightning itself defines no patches, but it probably _should_ be required in any Lightning projects that defines patches.

For instance, Lightning Core only has it as a dev dependency, which seems likely to cause problems if a consuming project hasn't required composer-patches from somewhere else.

Alternatively, Lightning just needs to heavily advertise / document that consuming projects need to `require cweagans/composer-patches` on their own.

danepowell (approved):

composer.json
TravisCarden: Probably the more _technically correct_ approach is for the components that declare patches to individually require `cweagans/composer-patches`. That seems safer than putting the requirement in the profile since components can be used à la carte, which is to say that we can't depend on the profile being present at all.

danepowell:
@phenaproxima I'd be really happy if you could merge this and cut a new release in the next day or two. The Drupal Scaffold conflict with Lightning is blocking a BLT 11 release (WIP at  https://github.com/acquia/blt/pull/3944). Let me know if I can do anything to help.

composer.json
phenaproxima: I agree with this. I've done this for a couple of components already and we will be able to continue doing it on a rolling basis. acquia/lightning-project already includes the patcher directly, so the current discrepancies should not affect actual users.

PR #666, opened 2019-12-19
[8.x-4.x] Stale configuration for views.view.frontpage.yml.

danepowell:
lightning_alter_frontpage_view() in lightning.profile dynamically modifies the views.view.frontpage.yml configuration at install time.

The problem is that it does so using an outdated Views configuration schema. Views has made some modifications recently such as [adding the `operator_limit_selection` key](https://www.drupal.org/node/2869168) and [removing the `core` key](https://www.drupal.org/node/3087832).

This causes configuration to appear perpetually overridden after an initial install, which is going to confuse people and cause problems for CI tools that expect a clean configuration state.

### DIFF
--- a/lightning.profile
+++ b/lightning.profile
@@ -139,7 +139,6 @@ function lightning_alter_frontpage_view() {
       ->set('tag', 'default')
       ->set('base_table', 'node_field_data')
       ->set('base_field', 'nid')
-      ->set('core', '8.x')
       ->set('display.default', [
         'display_options' => [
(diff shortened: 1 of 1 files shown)


phenaproxima:
Thanks, @danepowell. I'm in favor of this. Would you be willing to open a separate pull request for 8.x-3.x? It won't cherry-pick cleanly.

danepowell:
https://github.com/acquia/lightning/pull/667

PR #668, opened 2019-12-20
Fix type-o in GitHub url

bimsonz:
Does what it says on the tin...


phenaproxima:
Thank you! This is great, merging.

PR #669, opened 2020-01-03
composer scratchpad update

kdruckman:


phenaproxima:
I just rolled a stable release of Diff (8.x-1.0), which includes the deprecation fixes through 8.7.x. So we can remove that from this PR :)

tests/packages.yml
```diff
@@ -0,0 +1,23 @@
+---
+acquia/lightning:
```
phenaproxima, line 2: @TravisCarden: This file is also referenced in .travis.yml as ORCA_PACKAGES_CONFIG_ALTER. Will these changes have any ill side effects due to that?

TravisCarden: Actually, @phenaproxima, `tests/packages_alter.yml` is referenced by `ORCA_PACKAGES_CONFIG_ALTER`. Each file is used in only one case, so there should be problem.

phenaproxima: Ah, thanks! I missed that, appreciate you pointing that out. Carry on!

PR #670, opened 2020-01-30
Remove committed and conflicting core patch.

danepowell:
This patch was committed Jan 14: https://www.drupal.org/node/3096566

This causes any builds using a dev release of Drupal core to fail, such as BLT's HEAD tests. It's an allowed failure so not urgent, but it needs to get fixed at some point.

I guess this will need to wait until Drupal 8.8.2 is released.


drupal-org.make
```diff
@@ -53,19 +53,22 @@ projects[libraries][version] = 3.0-alpha1
 projects[lightning_api][type] = module
 projects[lightning_api][version] = 4.4
 projects[lightning_core][type] = module
-projects[lightning_core][version] = 5.0
+projects[lightning_core][download][type] = git
```
phenaproxima, line 56: Something doesn't make sense here. Why did this change? I'm surprised that it would have.

phenaproxima:
I fixed this in the 4.1.2 release (which was made specifically for this reason), and also figured out why Composer was insisting on installing 8.9.x-dev (fixed that too). Closing this one out :)

PR #671, opened 2020-03-05
Add a check for Libraries API module and warn users about the dependency being dropped

kdruckman:


lightning.install
```diff
@@ -9,6 +9,34 @@ use Drupal\Core\Entity\Entity\EntityViewDisplay;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\lightning_core\ConfigHelper as Config;
 
+/**
+ * Alerts users of Libraries API dependencies.
+ *
+ * A future version of Lightning will remove Libraries API, so
+ * users must add that dependency to composer.json manually.
+ */
```
phenaproxima, line 20: Let's add a `@todo Remove this function when Libraries API is no longer shipped with Lightning.` as well, so we remember to do it.

lightning.install
```diff
@@ -9,6 +9,34 @@ use Drupal\Core\Entity\Entity\EntityViewDisplay;
...
+function lightning_requirements() {
+    $requirements = [];
+    $libraries_dependencies = [];
+    $enabled_modules = \Drupal::moduleHandler()->getModuleList();
+    
+    foreach ($enabled_modules as $module => $data) {
```
phenaproxima: This can be a bit simpler; the module extension list has a `getAllInstalledInfo()` method which will return info for all installed modules.

lightning.install
```diff
@@ -9,6 +9,34 @@ use Drupal\Core\Entity\Entity\EntityViewDisplay;
...
+        $info = \Drupal::service('extension.list.module')->getExtensionInfo($module);
+        if (in_array('drupal:libraries', $info['dependencies'])) {
```
phenaproxima: Three things: we should always pass TRUE as a third argument to in_array(), the actual name of the project will be `libraries:libraries` rather than `drupal:libraries`, and we also need to support the old syntax (just `libraries`, with no project prefix).

lightning.install
```diff
@@ -9,6 +9,34 @@ use Drupal\Core\Entity\Entity\EntityViewDisplay;
...
+            $libraries_dependencies[$module] = $info['name'];
+        }
+    }
+
+    if ($enabled_modules) {
+        $requirements['lightning_libraries'] = [
+            'title' => t('Libraries API dependencies'),
+            'value' => t('These modules require Libraries API: %module_list. Lightning will soon remove Libraries API as a dependency, so you must add that dependency to your composer.json manually.', ['%module_list' => implode(', ', $libraries_dependencies)]),
```
phenaproxima: There's a nice helper function in Lightning Core that we can use to improve the formatting of the comma-separated list. See `\Drupal\lightning_core\Element::oxford()`. :)

phenaproxima (changes requested):
This is definitely what I was imagining! Nice work!! In addition to these changes, can we also add `drupal/libraries:^3.0` as an explicit dependency to Lightning's composer.json?

phenaproxima:
This looks pretty damn great! Since we've decided that Lightning Media will be the thing to deprecate and remove this dependency, let's close this PR and reconstitute it against Lightning Media 3.x.

lightning.install
```diff
@@ -8,6 +8,38 @@
...
+ *
+ * @todo Remove this function when Libraries API is no longer shipped with Lightning.
+ */
+function lightning_requirements() {
+  $requirements = [];
+  $libraries_dependencies = [];
+  $enabled_modules = \Drupal::service('extension.list.module')->getAllInstalledInfo();
+
+  foreach ($enabled_modules as $module => $info) {
+    foreach (['libraries:libraries', 'libraries'] as $needle) {
```
phenaproxima: We can remove a level of nesting here by using array_intersect(), like so:
```php
foreach ($enabled_modules as $module => $info) {
  if (array_intersect($info['dependencies'], ['libraries:libraries', 'libraries'])) {
    $libraries_dependencies[$module] = $info['name'];
  }
}
```

lightning.install
```diff
@@ -8,6 +8,38 @@
...
+      if (in_array($needle, $info['dependencies'], TRUE)) {
+        $libraries_dependencies[$module] = $info['name'];
+      }
+    }
+  }
+
+  if ($enabled_modules) {
```
phenaproxima: This should be `if ($libraries_dependencies)`. :)

lightning.install
```diff
@@ -8,6 +8,38 @@
...
+    $requirements['lightning_libraries'] = [
+      'title' => t('Libraries API dependencies'),
+      'value' => t('These modules require Libraries API: %module_list. Lightning will soon remove Libraries API as a dependency, so you must add that dependency to your composer.json manually.', ['%module_list' => Element::oxford($libraries_dependencies)]),
```
phenaproxima: Nit: To increase readability, can we stagger the translation variables like so:
```php
'value' => t('These modules...', [
  '%module_list' => Element::oxford($libraries_dependencies),
])
```

phenaproxima (changes requested):

lightning.install
kdruckman: yes, much better, thanks!

lightning.install
kdruckman: doh! Yes it should.

PR #673, opened 2020-03-24
Convert Drupal Console VersionCommand to a drush command.

kdruckman:


src/Commands/LightningCommands.php
```diff
@@ -1,19 +1,16 @@
...
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Yaml\Yaml;
 
 /**
- * Prints the semver version of the current Lightning code base.
+ * Exposes Drush commands provided by the Lightning profile.
  */
-class VersionCommand extends Command {
-
+class LightningCommands extends DrushCommands {
```
phenaproxima: Can we make this class final?

src/Commands/LightningCommands.php
```diff
@@ -22,28 +19,31 @@ class VersionCommand extends Command {
   protected $appRoot;
 
   /**
-   * VersionCommand constructor.
+   * LightningVersionCommand constructor.
```
phenaproxima: Wrong class name.

src/Commands/LightningCommands.php
```diff
@@ -22,28 +19,31 @@ class VersionCommand extends Command {
...
+  public function version() {
     $finder = (new Finder())
       ->files()
       ->name('lightning.info.yml')
       ->in($this->appRoot . '/profiles');
 
     foreach ($finder as $info_file) {
       $info = Yaml::parse($info_file->getContents());
-      if ($info['distribution']['name'] === 'Lightning' && isset($info['version'])) {
+      if ($info['name'] === 'Lightning' && isset($info['version'])) {
```
phenaproxima, line 44: Nice catch!

src/Commands/LightningCommands.php
```diff
@@ -54,9 +54,8 @@ protected function execute(InputInterface $input, OutputInterface $output) {
     if (!isset($info)) {
       throw new \Exception('Lightning info file not found.');
     }
-
-    $io = new DrupalStyle($input, $output);
-    $io->info($this->toSemanticVersion($info['version']));
+    $io = new DrushStyle($this->input(), $this->output());
```
phenaproxima: This might not be necessary. I think we might be able to just do `$this->output()->writeln('Version ' . static::toSemanticVersion(...))`.

src/Commands/LightningCommands.php
```diff
@@ -88,4 +87,4 @@ public static function toSemanticVersion($drupal_version) {
     return $semver;
   }
 
-}
+}
```
phenaproxima: Needs an empty line at the end of the file to appease git.

drush.services.yml
```diff
@@ -0,0 +1,7 @@
+services:
+  lightning.commands:
+    class: \Drupal\lightning\Commands\LightningCommands
+    arguments:
+      - '%app.root%'
+    tags:
+      - { name: drush.command }
```
phenaproxima: We need an empty line at the end of the file, for git reasons.

composer.json
```diff
@@ -102,6 +107,9 @@
             },
             "drupal/lightning_api": {
                 "#75 - Do not use deprecated path.alias_storage": "https://raw.githubusercontent.com/acquia/lightning/9.0.x-compatible/lightning_api-75.diff"
+            },
+            "drupal/metatag": {
```
phenaproxima, line 111: Is this needed?

phenaproxima (changes requested):
Minor changes needed, but looks great overall. Nice work!

composer.json
kdruckman: evidently. Weird.

phenaproxima (approved):

src/Commands/LightningCommands.php
```diff
@@ -77,7 +73,7 @@ protected function execute(InputInterface $input, OutputInterface $output) {
    *   The semantic version number.
    */
   public static function toSemanticVersion($drupal_version) {
-    preg_match('/^8\.x-(\d+).(\d)(\d+)(-.+)?$/', $drupal_version, $matches);
+    preg_match('/^9\.x-(\d+).(\d)(\d+)(-.+)?$/', $drupal_version, $matches);
```
phenaproxima: Let's modify this a bit -- change it to `/^[89]\.x` to allow either 8.x or 9.x here. That should fix the test which is currently failing.

phenaproxima (changes requested):
We're so close here. The dedicated unit test is currently failing, and I think it's due to the regex having changed without the test cases also having changed.

src/Commands/LightningCommands.php
kdruckman: Forgot to change that part. Thanks!

phenaproxima (approved):

phenaproxima:
The test failures are legitimate and expected, and not coming from this PR. I think this is ready -- merging it!

PR #675, opened 2020-03-30
Port Subprofile Generator to Drush command

kdruckman:


src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,55 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+
+class SubProfileGenerator extends BaseGenerator
+{
+
+  protected $name = 'lightning-subprofile';
+  protected $description = 'Generates a Lightning Subprofile.';
+  protected $alias = 'lsp';
```
phenaproxima, line 44: Nice idea!

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,55 @@
...
+   */
+  protected function interact(InputInterface $input, OutputInterface $output)
+  {
+    $questions['name'] = new Question('Profile Name');
+    $questions['name']->setValidator([Utils::class, 'validateRequired']);
+    $questions['machine_name'] = new Question('Profile Machine Name');
+    $questions['machine_name']->setValidator([Utils::class, 'validateMachineName']);
+    $questions['description'] = new Question('Enter the description (optional)', NULL);
+    $questions['install'] = new Question('Additional modules to include (optional), separated by commas (e.g. context, rules, file_entity)', NULL);
+    $questions['install']->setNormalizer(function ($answer) {
```
phenaproxima: Nice find!! I was wondering how to get questions to return arrays.

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,55 @@
...
+      return $answer ? array_map('trim', explode(',', $answer)) : [];
+    });
+    $questions['exclude'] = new Question('Lightning components to exclude (optional), separated by commas (e.g. lightning_workflow, lightning_media)', Null);
+    $questions['exclude']->setNormalizer(function ($answer) {
```
phenaproxima: We can reuse the normalizer function if we save it in a variable, like this:

```
$normalizer = function ($answer) {
  return $answer ? array_map('trim', explode(',', $answer)) : [];
};
$questions['install']->setNormalizer($normalizer);
$questions['exclude']->setNormalizer($normalizer);
```

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,55 @@
...
+      return $answer ? array_map('trim', explode(',', $answer)) : [];
+    });
+
+    $vars = &$this->collectVars($input, $output, $questions);
+
+    $this->addFile()
```
phenaproxima, line 132: I think we actually won't need to use a template for the info file at all. If we build an info array, I think we can do `$this->addFile()->path('...')->content(\Drupal\Component\Serialization\Yaml::encode($info_array));`

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,55 @@
...
+      ->path('contrib/{machine_name}/{machine_name}.info.yml')
```
phenaproxima: This should probably be `custom`, not `contrib`, maybe? Since the profile is a custom one we're generating.

phenaproxima (commented):
This is an AWESOME START!! Fantastic work! A few small comments here and there, but this is really promising.

src/Generators/SubProfileGenerator.php
kdruckman: yeah... this took me a while. :)

src/Generators/SubProfileGenerator.php
kdruckman: oh, that's a good point.

drush.services.yml
```diff
@@ -5,3 +5,10 @@ services:
       - '@app.root'
     tags:
       - { name: drush.command }
+
+  subprofile.generator:
+    class: Drupal\lightning\Generators\SubProfileGenerator
```
phenaproxima: Nit: Can we put single quotes around this class name?

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,120 @@
...
+use Drupal\lightning\ComponentDiscovery;
+use DrupalCodeGenerator\Command\BaseGenerator;
+use DrupalCodeGenerator\Utils;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+use Symfony\Component\Console\Question\Question;
+
+class SubProfileGenerator extends BaseGenerator
```
phenaproxima: This class needs a doc comment, and should be final.

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,120 @@
...
+{
+  /**
+   * The Lightning component discovery helper.
+   *
+   * @var \Drupal\lightning\ComponentDiscovery
+   */
+  protected $componentDiscovery;
```
phenaproxima: I don't mind continuing to use ComponentDiscovery for now (i.e., don't worry about fixing this right now), but we should eventually remove it. It was written aeons ago, when we thought we might have modules that don't start with `lightning_`. That never happened, though, so we could easily discover all Lightning modules by just filtering all modules from the `extension.list.module` service, looking for ones that start with `lightning_`.

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Modules to include in the sub-profile.
+   *
+   * @var string[]
+   */
+  protected $installList = [];
+
+  protected $name = 'lightning-subprofile';
```
phenaproxima, line 34: Nit: Can we add `{@inheritdoc}` comments above all these members that we're overriding from BaseGenerator?

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,120 @@
...
+   * {@inheritdoc}
+   */
+  protected function interact(InputInterface $input, OutputInterface $output)
+  {
+    $questions['name'] = new Question('Profile Name');
+    $questions['name']->setValidator([Utils::class, 'validateRequired']);
+
+    $questions['machine_name'] = new Question('Profile Machine Name (enter for default)');
+    $questions['machine_name']->setValidator([Utils::class, 'validateMachineName']);
+    $questions['description'] = new Question('Enter the description (optional)', NULL);
```
phenaproxima: Nit: Is NULL necessary? If so, a comment would be helpful here.

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,120 @@
...
+      );
+      $questions['exclude']->setMultiselect(true);
+      $this->collectVars($input, $output, $questions);
+    }
+
+
+    $info_array = [
+      'name' => $vars['name'],
+      'type' => 'profile',
+      'description' => [],
```
phenaproxima: Let's not put `description`, `install`, and `exclude` here to begin with if we can avoid it. We should only let them be added by the `if` statements below, so that they don't appear in the generated info file if they're not actually needed.

src/Generators/SubProfileGenerator.php
```diff
@@ -0,0 +1,120 @@
...
+    }
+    if ($vars['exclude']) {
+      $info_array['exclude'] = $vars['exclude'];
+    }
+
+    $info_array = array_filter($info_array);
+
+    $this->addFile()
+      ->path('custom/{machine_name}/{machine_name}.info.yml')
+      ->content(\Drupal\Component\Serialization\Yaml::encode($info_array));
```
phenaproxima: We should be `use` the Yaml class at the top of the file.

phenaproxima (changes requested):
My complaints are very minor (there are some additional coding standards things I didn't mention, but I can either fix those on commit, or we can run `phpcs` later). I think this is fabulous work and will be a lot better and easier to use than the Drupal Console equivalent.

src/Generators/SubProfileGenerator.php
kdruckman: This was the only way I could figure out how to get them in the right order.

src/Generators/SubProfileGenerator.php
kdruckman: oops :)

tests/src/Functional/SubprofileGeneratorTest.php
```diff
@@ -75,23 +90,25 @@ public function test(array $answers = []) {
 
     $answers += [
       'name' => 'Wizards',
-      'description' => 'This is the description.',
+      'machine_name' => 'wizards',
+      'description' => '',
```
kdruckman: This one has me stumped.

